PR #1, opened 2018-10-17
Initial skeleton

bison:
This an initial skeleton for the cluster-autoscaler-operator project. Still a ton of work to do, but it will create cluster-autoscaler pods with arguments based on the contents of a ClusterAutoscaler CRD.

Easiest way to run it for now is to start minikube and:
```sh-session
kubectl apply -f install/
export OPERATOR_NAME=cluster-autoscaler-operator
operator-sdk up local --namespace openshift-cluster-autoscaler
kubectl -n openshift-cluster-autoscaler apply -f examples/clusterautoscaler.yaml
```


examples/clusterautoscaler.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: "autoscaling.openshift.io/v1alpha1"
```
enxebre: In this scenario how would we define multiple machineSets? cc @frobware

frobware: You need multiple definitions, all uniquely named, with min and max bounds.

examples/clusterautoscaler.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: "ClusterAutoscaler"
+metadata:
+  name: "example"
+spec:
+  minReplicas: 1
```
frobware: Ah - The min and max replicas here is for the autoscaler itself. Yes?

examples/clusterautoscaler.yaml
enxebre: I think our CRD instance should represent a desired state of the autoscaler in the cluster, i.e I want this autoscaler to scale machineSets "a" with min/max 1/2 and machineSet "b" with min/max 2/3
Does it make sense?

frobware: When you say "express which ones" it's important to note that the CA is much more dynamic than that. Our clusterapi-based implementation will auto discover all machinesets (every 30s, configurable) and only discard those that do not have the following labels:

```
    sigs.k8s.io/cluster-api-autoscaler-node-group-min-size: "1"
    sigs.k8s.io/cluster-api-autoscaler-node-group-max-size: "4"
```

The min/max is a property of the machine set itself.

```
apiVersion: cluster.k8s.io/v1alpha1
kind: MachineSet
metadata:
  labels:
    sigs.k8s.io/cluster-api-cluster: amcdermo
    sigs.k8s.io/cluster-api-autoscaler-node-group-min-size: "3"
    sigs.k8s.io/cluster-api-autoscaler-node-group-max-size: "10"
  name: foo
```

```
apiVersion: cluster.k8s.io/v1alpha1
kind: MachineSet
metadata:
  labels:
    sigs.k8s.io/cluster-api-cluster: amcdermo
    sigs.k8s.io/cluster-api-autoscaler-node-group-min-size: "1"
    sigs.k8s.io/cluster-api-autoscaler-node-group-max-size: "40"
  name: bar
```

If you want a machine set that should not be used for autoscaling then don't add min/max labels. (Of course, we could express this in a different way. This is just how it is right now.)

enxebre: That sounds great. The use case I'm thinking is installer generates a given number of machineSets with min/max all based on user input. Then the autoscaler does the rest based on your comment above

frobware:
```
$ ./cluster-autoscaler --help 2>&1 | grep balance
--balance-similar-node-groups  -- Detect similar node groups and balance the number of nodes between them
```

I have tried that ^ on a small set of machinesets and it appears to do the right thing, however, the following is a default that I did not change, which I think we want to consider too:

```
$ ./cluster-autoscaler --help 2>&1 | grep expander
--expander string -- Type of node group expander to be used in scale up. Available values: [random,most-pods,least-waste,price] (default "random")
```

enxebre: cool thanks, those should be taken into consideration to model the CRD

bison: I'm glad you guys brought this up, because this is exactly where I left off -- trying to figure out how this bit should work.

My initial thought based on the example / prototype CRD was that you'd have multiple autoscaler pods each targeting a different MachineSet or other set of nodes. So if I have an "xxl-gpu-ssd" MachineSet, I could create a ClusterAutoscaler resource targeting that and give it its own autoscaler with unique settings. The operator could also add any annotations necessary to the MachineSet or nodes in addition to deploying the autoscaler.

If there's no way to target a specific MachineSet, and you're only expected to have a single autosclaer in a cluster, then I'm not sure there's much for this operator to actually do. Would it be any different than just using a deployment and a ConfigMap at that point?

examples/clusterautoscaler.yaml
bison: Actually was expecting this to represent that min and max for the target MachineSet. Just commented in the other thread with my confusion about the plan here though.

bison:
@frobware: I actually took the RBAC bits from upstream master as a starting place.

bison:
Merging this as a starting point. We can figure out the rest later.

PR #3, opened 2018-10-23
Add MachineAutoscaler resource

bison:
This is based on #2 and adds a rough first-pass at a MachineAutoscaler that modify labels on a target MachineSet object. Will rebase on master when the other PR is merged. Also adds build scripts copied from machine-api-operator.


pkg/autoscaler/handler.go
```diff
@@ -167,6 +220,39 @@ func asOwner(ca *v1alpha1.ClusterAutoscaler) metav1.OwnerReference {
...
+	}
+
+	return target, nil
+}
+
+func machineNeedsUpdate(have, want map[string]string) bool {
+	minDiff := have[minSizeLabel] != want[minSizeLabel]
+	maxDiff := have[maxSizeLabel] != want[maxSizeLabel]
+
+	if minDiff || maxDiff {
```
frobware: Just `return minDiff || maxDiff`?

bison: Done.

PR #8, opened 2018-11-01
Migrate to latest operator-sdk

bison:
This moves to the latest operator-sdk scaffolding, which uses a project layout based on the upstream controller-runtime. Everything looks a lot more like our other operators now.


bison:
/hold

This is still a WIP. Almost there though.

bison:
This is not going to pass yet. I'll re-run the tests when it's ready.

bison:
/hold cancel

This is ready.

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -0,0 +1,243 @@
...
+	// that reads objects from the cache and writes to the apiserver
+	client client.Client
+	scheme *runtime.Scheme
+}
+
+// Reconcile reads that state of the cluster for a ClusterAutoscaler
+// object and makes changes based on the state read and what is in the
+// ClusterAutoscaler.Spec
+func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log.Printf("Reconciling ClusterAutoscaler %s/%s\n", request.Namespace, request.Name)
```
enxebre: if I'm reading correctly we are not currently ensuring singleton autoscaler here but scoping the crd instance by namespace/name, so this would allow multiple autoscalers atm, correct?

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -0,0 +1,243 @@
...
+		return err
+	}
+
+	// Watch for changes to primary resource ClusterAutoscaler
+	err = c.Watch(&source.Kind{Type: &autoscalingv1alpha1.ClusterAutoscaler{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to secondary resources owned by a ClusterAutoscaler
```
enxebre: which ones are these?

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -0,0 +1,243 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	caImage          = "quay.io/bison/cluster-autoscaler:a554b4f5"
```
enxebre: we'll want to do this overridable via images configmap, env variable or similar

enxebre:
This is looking great! we should merge soon and proceed with smaller PRs. We'll need to provide integration testing to validate controllers deploy ca and set min/max labels and include an image-references file so this becomes part of the payload. For this to work "install/" manifests must exist into a folder called "manifests/" within the docker image

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
bison: Exactly. Wanted to get this in and do that bit next.

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
bison: The actual deployment object.

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
bison: Yeah, for sure. I have a TODO for that.

bison:
Thanks. Totally agree. Moving from v0.0.x to v0.1.0 was basically a rewrite, so this is kind of big unfortunately. If everything looks okay, we should merge and start tackling the other bits in separate PRs.

PR #9, opened 2018-11-06
Make ClusterAutoscaler resource a singleton

bison:
This makes the ClusterAutoscaler resource a singleton. The CRD becomes cluster-scoped and the operator only watches a single instance. The name of the instance to watch is configurable at runtime.

It also allows for overriding the cluster-autoscaler image via an environment variable.

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
@@ -35,8 +35,10 @@ const (
 // when the Manager is Started.
 func Add(mgr manager.Manager, cfg *operator.Config) error {
 	r := newReconciler(mgr)
+
 	r.SetImage(cfg.ClusterAutoscalerImage)
 	r.SetReplicas(cfg.ClusterAutoscalerReplicas)
+	r.SetNamespace(cfg.ClusterAutoscalerNamespace)
 
 	return add(mgr, cfg, r)
 }
@@ -116,6 +118,9 @@ type Reconciler struct {
 	client client.Client
 	scheme *runtime.Scheme
 
+	// The namespace for cluster-autoscaler deployments.
+	caNamespace string
+
 	// The cluster-autoscaler image to use in deployments.
 	caImage string
 
@@ -169,6 +174,11 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	return reconcile.Result{}, nil
 }
 
+// SetNamespace sets the cluster-autoscaler namespace on the reconciler.
+func (r *Reconciler) SetNamespace(namespace string) {
+	r.caNamespace = namespace
+}
+
 // SetImage sets the cluster-autoscaler image on the reconciler.
 func (r *Reconciler) SetImage(image string) {
 	r.caImage = image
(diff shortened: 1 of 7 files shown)


bison:
/hold

Still testing this out a bit.

bison:
/hold cancel

I think this is ready now.

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -17,43 +17,76 @@ import (
...
 	// Create a new controller
 	c, err := controller.New("clusterautoscaler-controller", mgr, controller.Options{Reconciler: r})
 	if err != nil {
 		return err
 	}
 
+	// ClusterAutoscaler is effectively a singleton resource.  A
+	// deployment is only created if an instance is found matching the
+	// name set at runtime.
+	p := predicate.Funcs{
```
enxebre: don't we want to include delete?

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -17,43 +17,76 @@ import (
...
 	}
 }
 
+// processEvent is used in predicate functions.  It inspects metadata and
+// returns a boolean value representing whether the controller should process an
+// event for the object.
+func processEvent(meta metav1.Object, cfg *operator.Config) bool {
+	// Only process events for objects matching the configured resource name.
+	if meta.GetName() != cfg.ClusterAutoscalerName {
+		glog.Warningf("Not processing ClusterAutoscaler %s", meta.GetName())
```
enxebre: We should probably indicate that only `config.ClusterAutoscalerName` will be watched

enxebre: I wonder if we could somehow set the name programatically so no need to arbitrarily having to match the manifests and the config names, I don't have any good suggestion atm though

enxebre:
looks great

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
enxebre: unit test for `processEvent` would be nice

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
bison: Delete doesn't do anything here. We rely on garbage collection and proper owner references to clean up resources. We could add it for completeness.

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
bison: I don't think we can do anything other than filter on names like this. This is honestly pretty weird. These kinds of singleton resources that are managing applications that are *already* native Kubernetes apps have dubious benefits. In a perfect world, the cluster-autoscaler itself would just be doing this. There would be no need for an operator.

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
bison: Added delete and generic predicates here.

PR #10, opened 2018-11-07
Add initial unit-test suite

bison:
This adds the start of a unit-test suite. It also sets version information at build time, and adds a type to help with generating cluster-autoscaler arguments.


enxebre:
nice
/lgtm

PR #13, opened 2018-11-19
Use annotations instead of labels for min and max

bison:
The cluster-autoscaler has moved to using annotations instead of
labels for the min and max sizes of a group.  This updates
MachineTarget resources to set the limits as annotations.

  https://github.com/openshift/kubernetes-autoscaler/pull/11

Note: This doesn't perform any kind of migration from labels to annotations. I don't think anyone is using this yet, or is even really using the cluster-api provider in the autoscaler, so it's probably not necessary. We could do it if we wanted though.

### DIFF
--- a/pkg/controller/machineautoscaler/machinetarget.go
+++ b/pkg/controller/machineautoscaler/machinetarget.go
@@ -8,9 +8,9 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-// ErrTargetMissingLabels is the error returned when a target is
-// missing the min or max labels.
-var ErrTargetMissingLabels = errors.New("missing min or max label")
+// ErrTargetMissingAnnotations is the error returned when a target is
+// missing the min or max annotations.
+var ErrTargetMissingAnnotations = errors.New("missing min or max annotation")
 
 // MachineTarget represents an unstructured target object for a
 // MachineAutoscaler, used to update metadata only.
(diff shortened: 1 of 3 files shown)


pkg/controller/machineautoscaler/machinetarget_test.go
```diff
@@ -80,35 +80,35 @@ func TestSetLimits(t *testing.T) {
...
 
 	_, _, err = target.GetLimits()
 	if err == nil {
-		t.Fatal("expected bad label error")
+		t.Fatal("expected bad annotation error")
 	}
 
-	// Set correct labels.
+	// Set correct annotations.
 	expectedMin, expectedMax := 2, 4
```
frobware: Do we want to test for (only) positive numbers too?

pkg/controller/machineautoscaler/machinetarget.go
```diff
@@ -33,50 +33,50 @@ func (mt *MachineTarget) NeedsUpdate(min, max int) bool {
...
 
 	if !minOK || !maxOK {
-		return 0, 0, ErrTargetMissingLabels
+		return 0, 0, ErrTargetMissingAnnotations
 	}
 
 	min, err = strconv.Atoi(minString)
 	if err != nil {
-		return 0, 0, fmt.Errorf("bad min label: %s", minString)
+		return 0, 0, fmt.Errorf("bad min annotation: %s", minString)
```
frobware: Perhaps we could convey the annotation key, might make it easier for someone to debug as there would be something concrete to grep for:
 
    return 0, 0, fmt.Errorf("bad min annotation for %q: %s", minSizeAnnotation, minString)

or something like that.

bison:
@frobware Happy to add both the things you mentioned in a follow up. I want to add validation to the CRD definitions soon as well.

PR #19, opened 2018-12-04
Run autoscaler pods on master nodes

### LINKED ISSUE #18: cluster autoscaler should always run on master machines
current operator is allowing deployment to worker nodes.

this appears to cause problems on scale down.

bison:
Adds the toleration and node selector to the autoscaler pod to schedule it on master nodes.

Fixes #18.


bison:
The command in the `test-e2e` make target worked on go-1.11, but not 1.10 -- which was causing the build to fail. Fixed it.

PR #22, opened 2018-12-13
Refactor: Move manager setup to operator package

bison:
This moves the set-up of the controller-manager and the reconcilers to
the operator package.  This will make it easier to add additional
supporting functions like reporting status to the CVO.


cmd/manager/main.go
```diff
@@ -25,62 +18,19 @@ func printVersion() {
...
-
-	// Setup Scheme for all resources
-	if err := apis.AddToScheme(mgr.GetScheme()); err != nil {
-		glog.Fatal(err)
-	}
-
-	// Setup all Controllers
-	if err := controller.AddToManager(mgr, operatorConfig); err != nil {
+	glog.Info("Starting cluster-autoscaler-operator")
+	if err := operator.Start(); err != nil {
```
frobware: This seems like a much better way to both start and exit cleanly. +1

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -78,16 +75,16 @@ func add(mgr manager.Manager, cfg *operator.Config, r reconcile.Reconciler) erro
 	// name set at runtime.
 	p := predicate.Funcs{
 		CreateFunc: func(e event.CreateEvent) bool {
-			return processEvent(e.Meta, cfg)
+			return r.NamePredicate(e.Meta)
```
frobware: nit: it might read a little better if it was something like `r.NamesMatch)` - just seems more positive.

frobware: Also, if nothing else uses `NamedPredicate()` you could just inline it here for locality: 

    predicate := func () ...

PR #24, opened 2018-12-18
operator: Add optional leader-election on startup

bison:
Adds leader-election on startup using a ConfigMap.

### DIFF
--- a/pkg/operator/config.go
+++ b/pkg/operator/config.go
@@ -1,11 +1,28 @@
 package operator
 
-import "os"
+import (
+	"os"
+	"strconv"
+
+	"github.com/golang/glog"
+)
 
 const (
 	// DefaultWatchNamespace is the default namespace the operator
 	// will watch for instances of its custom resources.
 	DefaultWatchNamespace = "openshift-cluster-api"
+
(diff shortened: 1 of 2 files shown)


pkg/operator/config.go
```diff
@@ -67,6 +97,24 @@ func ConfigFromEnvironment() *Config {
 		config.WatchNamespace = watchNamespace
 	}
 
+	if leaderElection, ok := os.LookupEnv("LEADER_ELECTION"); ok {
+		le, err := strconv.ParseBool(leaderElection)
+		if err != nil {
+			le = DefaultLeaderElection
+			glog.Errorf("Error parsing LEADER_ELECTION environment variable: %v", err)
```
frobware: Is the intention here that this falls through and `le` is (possibly) `false` (depending on implementation)?

bison: If the environment variable is present, but fails to parse, we log the error and use the default value. That happens to be `true` at the moment.

frobware: Ah, sorry. Missed the `le = DefaultLeaderElection`. Somehow!

PR #25, opened 2018-12-20
Mark targets as owned by MachineAutoscaler resources

bison:
This adds an annotation marking targets as owned by the
MachineAutoscaler resources that reference them.  It also watches all
supported target types for changes and queues reconciliation requests
for the owning MachineAutoscaler resource.  If multiple
MachineAutoscaler resources reference the same target, the first wins
and reconciliation will fail for the others.

### DIFF
--- a/pkg/controller/machineautoscaler/machinetarget.go
+++ b/pkg/controller/machineautoscaler/machinetarget.go
@@ -4,13 +4,53 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"strings"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 )
 
-// ErrTargetMissingAnnotations is the error returned when a target is
-// missing the min or max annotations.
-var ErrTargetMissingAnnotations = errors.New("missing min or max annotation")
+var (
+	// ErrTargetMissingAnnotations is the error returned when a target is
+	// missing the min or max annotations.
+	ErrTargetMissingAnnotations = errors.New("missing min or max annotation")
+
+	// ErrTargetAlreadyOwned is the error returned when a target is already
+	// marked as owned by another MachineAutoscaler resource.
+	ErrTargetAlreadyOwned = errors.New("already owned by another MachineAutoscaler")
+
+	// ErrTargetMissingOwner is the error returned when a target has no owner
+	// annotations set.
+	ErrTargetMissingOwner = errors.New("missing owner annotation")
+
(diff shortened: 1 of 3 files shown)


pkg/controller/machineautoscaler/machinetarget.go
```diff
@@ -81,3 +121,68 @@ func (mt *MachineTarget) GetLimits() (min, max int, err error) {
...
+	return true, nil
+}
+
+// RemoveOwner removes the owner annotation from the target.
+func (mt *MachineTarget) RemoveOwner() {
+	annotations := mt.GetAnnotations()
+
+	delete(annotations, MachineTargetOwnerAnnotation)
+
+	mt.SetAnnotations(annotations)
```
frobware: At face value the call to `Setannotations` looks redundant. We Get, then Set without appearing to change anything - curious as to why?

pkg/controller/machineautoscaler/machinetarget_test.go
```diff
@@ -138,3 +143,87 @@ func TestRemoveLimits(t *testing.T) {
...
+		t.Fatalf("failed to get owner: %v", err)
+	}
+
+	if nn.Name != "owner" || nn.Namespace != "owner" {
+		t.Error("target returned unexpected owner")
+	}
+
+	// Malformed owner.
+	target.SetAnnotations(map[string]string{
+		MachineTargetOwnerAnnotation: "to/many/parts/here",
```
frobware: nit: s/to/too

pkg/controller/machineautoscaler/machinetarget.go
bison: We delete the owner annotation. So we get the annotations map, remove the owner annotation, then set the resulting map with no owner back on the target.

pkg/controller/machineautoscaler/machinetarget_test.go
bison: Fixed.

enxebre:
great! It'd be nice to start to add e2e coverage for things like this

pkg/controller/machineautoscaler/machinetarget_test.go
```diff
@@ -5,35 +5,40 @@ import (
...
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 }
 
-// NewTarget returns a new MachineTarget referencing an TestObject.
-func NewTarget() *MachineTarget {
-	obj := &TestObject{
+// NewTargetOnwer returns a new TargetOwner with the given name and
+// namespace set.
+func NewTargetOnwer(namespace, name string) *TargetOwner {
```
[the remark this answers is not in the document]
bison: Oops. If the current test run passes, I'll push a fix for this. Otherwise I'll fix it in a follow up PR.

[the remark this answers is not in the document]
bison: Fixed here.

PR #31, opened 2019-01-18
Finalize previous MachineAutoscaler targets on update

bison:
This records the most recently observed target of a MachineAutoscaler in the status.  This is used to detect an update to the target and remove autoscaling configuration from the previously targeted resource before applying configuration to the new target.


bison:
/hold

This is a WIP. Needs testing.

pkg/controller/machineautoscaler/machinetarget.go
```diff
@@ -186,3 +186,31 @@ func (mt *MachineTarget) GetOwner() (types.NamespacedName, error) {
 
 	return nn, nil
 }
+
+// RemoveAnnotations removes the annotations with the given keys from the
+// target.  It returns a bool indicating whether the annotations were actually
```
frobware: Just curious about how the caller distinguishes between found/not found when multiple annotations are passed but only some are deleted.

bison: I'm not really sure this is a great API... We can change it at some point. In any case, this is really just meant to signal that the annotations changed in some way -- nothing more specific than that. It makes it easier to know if you need to update the resource with the API or not.

pkg/controller/machineautoscaler/machineautoscaler_controller.go
```diff
@@ -43,11 +44,45 @@ var SupportedTargetGVKs = []schema.GroupVersionKind{
...
 
+// Config represents the configuration for a reconciler instance.
+type Config struct {
+	// The namespace for MachineAutosclaers and their targets.
+	Namespace string
+}
+
+// ObjectReference represents a CrossVersionObjectReference
+// identifying a target resource with additional helper methods.
+type ObjectReference v1alpha1.CrossVersionObjectReference
```
enxebre: Coudn't we use https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/objectreference.go?

bison: Yep. Missed that. Done.

PR #32, opened 2019-01-24
Properly validate MachineAutoscaler target references

bison:
This adds validation to MachineAutoscaler target references.
Previously, if for example a reference had no name set, the dynamic
client would return a list and cause a panic.  This prevents fetching
invalid or unsupported target references.

### DIFF
--- a/pkg/controller/machineautoscaler/machineautoscaler_controller.go
+++ b/pkg/controller/machineautoscaler/machineautoscaler_controller.go
@@ -34,9 +34,15 @@ const (
 	maxSizeAnnotation = "sigs.k8s.io/cluster-api-autoscaler-node-group-max-size"
 )
 
-// ErrUnsupportedTarget is the error returned when a target references an object
-// with an unsupported GroupVersionKind.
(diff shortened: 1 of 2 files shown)


pkg/controller/machineautoscaler/machineautoscaler_controller.go
```diff
@@ -228,8 +234,8 @@ func (r *Reconciler) GetTarget(ref *corev1.ObjectReference) (*MachineTarget, err
 	obj := &unstructured.Unstructured{}
 	gvk := ref.GroupVersionKind()
 
-	if !SupportedTarget(gvk) {
-		return nil, ErrUnsupportedTarget
+	if valid, err := ValidateReference(ref); !valid || err != nil {
```
frobware: Is it more prudent to first look at `err`, then `!valid`.

bison: Yeah, this is a little weird. I've updated the comment for the validation function to reflect the assumptions about returning non-nil errors. I also updated the test to ensure we always return a non-nil error for invalid references.

PR #34, opened 2019-02-01
Use the correct ServiceAccount for the operator

bison:
The role and binding were present in the CVO manifests, but the
actual ServiceAccount was missing.  This adds it, and configures the
operator deployment to use it.

### DIFF
--- a/install/0000_50_cluster-autoscaler-operator_03_rbac.yaml
+++ b/install/0000_50_cluster-autoscaler-operator_03_rbac.yaml
@@ -1,4 +1,17 @@
 ---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cluster-autoscaler-operator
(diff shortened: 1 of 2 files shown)


install/0000_50_cluster-autoscaler-operator_03_rbac.yaml
```diff
@@ -32,6 +45,16 @@ rules:
   - statefulsets
   verbs:
   - "*"
+- apiGroups:
+  - cluster.k8s.io
+  resources:
+  - machinedeployments
```
ingvagabund: +1 even though not yet supported

ingvagabund:
Don't think `cluster-autoscaler-operator` needs permissions to updated endpoints. Other resources and verbs might need to be revisit. Let's do it after we deploy the kubemark actuator smoke testing. Good for now.

/lgtm

PR #35, opened 2019-02-04
Update Cluster autoscaler operator to consume machine.openshift.io

spangenberg:


spangenberg:
> ```
> minSizeAnnotation = "sigs.k8s.io/cluster-api-autoscaler-node-group-min-size"
> maxSizeAnnotation = "sigs.k8s.io/cluster-api-autoscaler-node-group-max-size"
> ```
> Both of these need to become `"machine.openshift.io/..."`

Thanks for the catch.

install/0000_50_cluster-autoscaler-operator_03_rbac.yaml
```diff
@@ -47,6 +47,7 @@ rules:
   - "*"
 - apiGroups:
   - cluster.k8s.io
+  - machine.openshift.io
```
frobware: Don't we need `machine.openshift.io` here too?

https://github.com/openshift/cluster-autoscaler-operator/blob/39f0bee0f2c0108aa02d383c31f45ec1bf4a5987/install/0000_50_cluster-autoscaler-operator_03_rbac.yaml#L149

bison: Oops. Yes.

pkg/controller/machineautoscaler/machineautoscaler_controller.go
```diff
@@ -47,8 +47,10 @@ var (
 // SupportedTargetGVKs is the list of GroupVersionKinds supported as targets for
 // a MachineAutocaler instance.
 var SupportedTargetGVKs = []schema.GroupVersionKind{
-	{Group: "cluster.k8s.io", Version: "v1alpha1", Kind: "MachineSet"},
 	{Group: "cluster.k8s.io", Version: "v1alpha1", Kind: "MachineDeployment"},
+	{Group: "cluster.k8s.io", Version: "v1alpha1", Kind: "MachineSet"},
+	{Group: "machine.openshift.io", Version: "v1beta1", Kind: "MachineDeployment"},
```
enxebre: Not a blocker, this would belong to a different PR

frobware: We need to decide quickly. this will merge RSN if the tests pass as we have LGTM and APPROVE.

bison: What belongs to a different PR? Just adding `MachineDeployments`? Or am I misunderstanding?

There's no harm in adding the deployments here.

bison:
/test e2e-aws-operator

Tests are passing for me locally. Not sure what was up on the last run.

enxebre:
`F0207 11:15:21.543907       1 cloud_provider_builder.go:90] Unknown cloud provider: openshift-machine-api`
can you either pinning the autoscaler image on cao (we'd need to keep both groups here https://github.com/openshift/cluster-autoscaler-operator/pull/35/files#diff-defb693cda2c7b43fb8c51ddc5cf70deL33) or just disable the deployment test as part of this PR, then re enable it after https://github.com/openshift/kubernetes-autoscaler/pull/29 gets merged

test/e2e/main.go
```diff
@@ -67,11 +67,12 @@ func runSuite() error {
 	}
 	glog.Info("PASS: CreateClusterAutoscaler")
 
-	if err := ExpectClusterAutoscalerAvailable(); err != nil {
-		glog.Errorf("FAIL: ExpectClusterAutoscalerAvailable: %v", err)
-		return err
-	}
-	glog.Info("PASS: ExpectClusterAutoscalerAvailable")
+	// TODO: Disabled temporarily
```
enxebre: can you add some context on why to the TODO?

spangenberg: Should be good now.

PR #36, opened 2019-02-06
Use kustomize to deploy CA operator in k8s

ingvagabund:
SSIA

### DIFF
--- a/kustomization.yaml
+++ b/kustomization.yaml
@@ -1,0 +1,13 @@
+# Adds namespace to all resources.
+namespace: kube-system
+
+# Each entry in this list must resolve to an existing
+# resource definition in YAML.  These are the resource
+# files that kustomize reads, modifies and emits as a
+# YAML string, with resources separated by document
+# markers ("---").
+resources:
+- install/0000_50_cluster-autoscaler-operator_01_clusterautoscaler.crd.yaml
+- install/0000_50_cluster-autoscaler-operator_02_machineautoscaler.crd.yaml
+- install/0000_50_cluster-autoscaler-operator_03_rbac.yaml
+- install/0000_50_cluster-autoscaler-operator_04_deployment.yaml
--- a/install/0000_50_cluster-autoscaler-operator_04_deployment.yaml
+++ b/install/0000_50_cluster-autoscaler-operator_04_deployment.yaml
@@ -33,6 +33,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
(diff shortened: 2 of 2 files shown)


pkg/apis/autoscaling/v1alpha1/clusterautoscaler_types.go
```diff
@@ -64,4 +64,5 @@ type ScaleDownConfig struct {
 	DelayAfterAdd     string `json:"delayAfterAdd"`
 	DelayAfterDelete  string `json:"delayAfterDelete"`
 	DelayAfterFailure string `json:"delayAfterFailure"`
+	UnneededTime      string `json:"unneededTime"`
```
frobware: I think the variable should reflect `scale-down`. And it should be optional too.

pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -34,6 +34,7 @@ func (a AutoscalerArg) TypeRange(t string, min, max int) string {
 // operator when processing ClusterAutoscaler resources.
 const (
 	LogToStderrArg                  AutoscalerArg = "--logtostderr"
+	VerboseArg                      AutoscalerArg = "--v=3"
```
frobware: I've typically found `4` to be more useful.

pkg/apis/autoscaling/v1alpha1/clusterautoscaler_types.go
frobware: Ah, but this is in `ScaleDownConfig`... ok, forgot my comment.

ingvagabund:
Just verified I am able to scale down pretty fast after applying the PR.

pkg/apis/autoscaling/v1alpha1/clusterautoscaler_types.go
ingvagabund: > I think the variable should reflect `scale-down`

I what sense? It's under `ScaleDownConfig` struct.

> And it should be optional too.

Yeah. Agree.

kustomization.yaml
```diff
@@ -0,0 +1,24 @@
+# Adds namespace to all resources.
+namespace: kube-system
+
+# Value of this field is prepended to the
```
frobware: If we are not using any of these keys then why not delete them all? Without them I don't have to guess whether I need them or not.

ingvagabund: Removed

enxebre:
/hold

PR #37, opened 2019-02-07
Allow to configure --scale-down-unneeded-time

ingvagabund:
It takes 10 minutes before a cluster is scalled down (default value of --scale-down-unneeded-time).
Allowing to override the option allows to decrease the time spend waiting
on scale down operation in both testing and development environment.

Also can help customer to scale down sooner.

### DIFF
--- a/pkg/apis/autoscaling/v1alpha1/clusterautoscaler_types.go
+++ b/pkg/apis/autoscaling/v1alpha1/clusterautoscaler_types.go
@@ -64,4 +64,5 @@ type ScaleDownConfig struct {
 	DelayAfterAdd     string `json:"delayAfterAdd"`
 	DelayAfterDelete  string `json:"delayAfterDelete"`
 	DelayAfterFailure string `json:"delayAfterFailure"`
+	UnneededTime      string `json:"unneededTime,omitempty"`
 }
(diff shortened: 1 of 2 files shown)


pkg/apis/autoscaling/v1alpha1/clusterautoscaler_types.go
```diff
@@ -64,4 +64,5 @@ type ScaleDownConfig struct {
 	DelayAfterAdd     string `json:"delayAfterAdd"`
 	DelayAfterDelete  string `json:"delayAfterDelete"`
 	DelayAfterFailure string `json:"delayAfterFailure"`
+	UnneededTime      string `json:"unneededTime"`
```
frobware: Should this be optional?

examples/clusterautoscaler.yaml
```diff
@@ -25,3 +25,4 @@ spec:
     delayAfterAdd: 10s
     delayAfterDelete: 10s
     delayAfterFailure: 10s
+    unneededTime: 10s
```
frobware: Not sure this should be the default. Machines will disappear in a production setting far quicker than they should.

ingvagabund: Ha, this is not supposed to be here. Thanks.

pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -96,6 +97,10 @@ func ScaleDownArgs(sd *v1alpha1.ScaleDownConfig) []string {
 		ScaleDownDelayAfterFailureArg.Value(sd.DelayAfterFailure),
 	}
 
+	if sd.UnneededTime != "" {
```
frobware: Let's also add a unit test case that validates a config with/without this value.

bison: We should actually probably add `omitEmpty` to all of the scale down argument fields in the resource, or even make them pointers. Then fix the `Value()` method here to return an empty string if it gets `nil` or an empty string. Not sure if we do that here or in a follow up.

frobware: I would vote in a follow-up.

enxebre:
/hold

PR #38, opened 2019-02-07
More verbose

ingvagabund:
Be more verbose to see more logs to improve debugging capability.

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler.go
@@ -34,6 +34,7 @@ func (a AutoscalerArg) TypeRange(t string, min, max int) string {
 // operator when processing ClusterAutoscaler resources.
 const (
 	LogToStderrArg                  AutoscalerArg = "--logtostderr"
+	VerboseArg                      AutoscalerArg = "--v=3"
 	NamespaceArg                    AutoscalerArg = "--namespace"
 	CloudProviderArg                AutoscalerArg = "--cloud-provider"
 	MaxGracefulTerminationSecArg    AutoscalerArg = "--max-graceful-termination-sec"
@@ -56,6 +57,7 @@ func AutoscalerArgs(ca *v1alpha1.ClusterAutoscaler, namespace string) []string {
 
 	args := []string{
 		LogToStderrArg.String(),
+		VerboseArg.String(),
 		CloudProviderArg.Value("cluster-api"),
 		NamespaceArg.Value(namespace),
 	}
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -34,6 +34,7 @@ func (a AutoscalerArg) TypeRange(t string, min, max int) string {
 // operator when processing ClusterAutoscaler resources.
 const (
 	LogToStderrArg                  AutoscalerArg = "--logtostderr"
+	VerboseArg                      AutoscalerArg = "--v=4"
```
frobware: I think it is one thing to expose the capability but I'm not sure we should default to such a high level.

ingvagabund: https://github.com/openshift/cluster-autoscaler-operator/pull/36#discussion_r254350896 huh?

ingvagabund: I can go back to 3 as I was

frobware: > [#36 (comment)](https://github.com/openshift/cluster-autoscaler-operator/pull/36#discussion_r254350896) huh?

:) so what I really mean is that if you're trying to debug then 4 yields information that has typically helped me debug - largely because any of the cluster-api cloudprovider logging is at level 4. 

But OOTB I'm asking if we want to default to level 4?

ingvagabund: The verbose level is set to 3 in many places. Sticking with that here as well.

> But OOTB I'm asking if we want to default to level 4?

I don't think so

pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -56,6 +57,7 @@ func AutoscalerArgs(ca *v1alpha1.ClusterAutoscaler, namespace string) []string {
 
 	args := []string{
 		LogToStderrArg.String(),
+		VerboseArg.String(),
 		CloudProviderArg.Value("cluster-api"),
```
frobware: Does this PR need rebasing? This is `openshift-machine-api` now.

pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -34,6 +34,7 @@ func (a AutoscalerArg) TypeRange(t string, min, max int) string {
 // operator when processing ClusterAutoscaler resources.
 const (
 	LogToStderrArg                  AutoscalerArg = "--logtostderr"
+	VerboseArg                      AutoscalerArg = "--v=3"
```
bison: This should not have a value. It should just be `--v` and then `VerboseArg.Value("3")` can be used to get this level. I would actually prefer a `CLUSTER_AUTOSCALER_VERBOSITY` setting that set this, but we don't have to do that now.

enxebre:
/hold

bison:
Closing in favor of https://github.com/openshift/cluster-autoscaler-operator/pull/53.

PR #39, opened 2019-02-07
Dump cluster autoscaler logs

ingvagabund:
Dump logs from the cluster operator pod so we have some idea of what went wrong in case the underlying container fails.


bison:
OpenShift CI already records these logs for us. Is this intended to be used somewhere else?

PR #44, opened 2019-02-08
Remove unregistered types from supported target types

bison:
This checks for errors resulting from unregistered types when
initializing the MachineAutoscaler reconciler and removes them from
the list of supported target types.

### DIFF
--- a/pkg/controller/machineautoscaler/machineautoscaler_controller.go
+++ b/pkg/controller/machineautoscaler/machineautoscaler_controller.go
@@ -10,6 +10,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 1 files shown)


ingvagabund:
This might hide the case where a user forgets to deploy machineset crd. Worth documenting it somewhere. Or, maybe to make at least one of machineset or machinedeployment crds exists. If not, fail?

bison:
> This might hide the case where a user forgets to deploy machineset crd. Worth documenting it somewhere. Or, maybe to make at least one of machineset or machinedeployment crds exists. If not, fail?

I don't want to try to do anything too clever, but we could check that there is at least one supported type found otherwise return an error. Does that sound like a reasonable compromise?

ingvagabund:
> I don't want to try to do anything too clever, but we could check that there is at least one supported type found otherwise return an error. Does that sound like a reasonable compromise?

+1

pkg/controller/machineautoscaler/machineautoscaler_controller.go
```diff
@@ -94,11 +99,23 @@ func (r *Reconciler) AddToManager(mgr manager.Manager) error {
 				ToRequests: handler.ToRequestsFunc(targetOwnerRequest),
 			})
 
-		if err != nil {
+		// If we get an error indicating that no matching type is registered
+		// with the API, then remove it from the list of supported target GVKs.
+		// If the type is later registered, a restart of the operator will pick
```
frobware: Does this have to be a forced restart (by a human)?

bison: Yes. For now. Seems reasonable to me. The immediate use case for this is just to not crash if one of the upstream or OpenShift groups is missing.

enxebre:
I agree with @bison is out of the scope to deploy SupportedTargetGVKs list
/approve

PR #47, opened 2019-02-11
Prepare test/e2e to be run over pure k8s (over kubemark based machine API)

ingvagabund:
We need to be able to iterate quickly and test scenarios without
consuming too much resouces.

### DIFF
--- a/test/e2e/main.go
+++ b/test/e2e/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"flag"
+	"strings"
 
 	"github.com/golang/glog"
 	"github.com/openshift/cluster-autoscaler-operator/pkg/apis"
@@ -11,9 +12,16 @@ import (
 )
 
 const (
-	namespace = "openshift-machine-api"
-	caName    = "default"
+	caName = "default"
 )
+
+var focus string
+var namespace string
+
+func init() {
+	flag.StringVar(&focus, "focus", "[openshift]", "If set, run only tests containing focus string. E.g. [k8s]")
+	flag.StringVar(&namespace, "namespace", "openshift-machine-api", "cluster-autoscaler-operator namespace")
+}
 
 var F *Framework
 
@@ -55,23 +63,40 @@ func main() {
 }
 
 func runSuite() error {
-	if err := ExpectOperatorAvailable(); err != nil {
-		glog.Errorf("FAIL: ExpectOperatorAvailable: %v", err)
-		return err
-	}
-	glog.Info("PASS: ExpectOperatorAvailable")
 
-	if err := CreateClusterAutoscaler(); err != nil {
-		glog.Errorf("FAIL: CreateClusterAutoscaler: %v", err)
-		return err
+	expectations := []struct {
+		expect func() error
+		name   string
+	}{
+		{
+			expect: ExpectOperatorAvailable,
+			name:   "[k8s][openshift] Expect operator to be available",
+		},
+		{
+			expect: CreateClusterAutoscaler,
+			name:   "[openshift] Create Cluster Autoscaler resource",
+		},
+		{
+			expect: ExpectClusterAutoscalerAvailable,
+			name:   "[openshift] Expect Cluster Autoscaler available",
+		},
+		{
+			expect: ExpectToScaleUpAndDown,
+			name:   "[k8s] Expect to scale up and down",
+		},
 	}
-	glog.Info("PASS: CreateClusterAutoscaler")
 
-	if err := ExpectClusterAutoscalerAvailable(); err != nil {
-		glog.Errorf("FAIL: ExpectClusterAutoscalerAvailable: %v", err)
-		return err
+	for _, tc := range expectations {
+		if strings.HasPrefix(tc.name, focus) {
+			if err := tc.expect(); err != nil {
+				glog.Errorf("FAIL: %v: %v", tc.name, err)
+				return err
+			}
+			glog.Infof("PASS: %v", tc.name)
+		} else {
+			glog.Infof("SKIPPING: %v", tc.name)
+		}
 	}
-	glog.Info("PASS: ExpectClusterAutoscalerAvailable")
 
 	return nil
 }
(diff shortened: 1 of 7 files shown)


bison:
Looks good. Small nit, I'd prefer to keep stuff like `cluster-autoscaler-operator-patch.yaml` and `kustomization.yaml` out of the root of the repo and in a directory like `hack/` or `test/`. Obviously not a big deal though.

ingvagabund:
Just rebasing

ingvagabund:
> Looks good. Small nit, I'd prefer to keep stuff like `cluster-autoscaler-operator-patch.yaml` and `kustomization.yaml` out of the root of the repo and in a directory like `hack/` or `test/`. Obviously not a big deal though.

I will address it eventually. Still a lot of moving pieces. I will move more shell script in the CI into this repo and move the kustomization.yaml file and others under more appropriate place.

bison:
Cool! No problem. Still LGTM.

test/e2e/main.go
```diff
@@ -11,10 +12,17 @@ import (
...
-	caName    = "default"
+	caName = "default"
 )
 
+var focus string
+var namespace string
+
+func init() {
+	flag.StringVar(&focus, "focus", "[openshift]", "If set, run only tests containing focus string. E.g. [k8s]")
+	flag.StringVar(&namespace, "namespace", "openshift-machine-api", "cluster API operator namespace")
```
enxebre: "cluster autoscaler operator namespace" ?

ingvagabund:
@enxebre comment address PTAL

test/e2e/main.go
```diff
@@ -55,23 +63,40 @@ func main() {
...
+	expectations := []struct {
+		expect func() error
+		name   string
+	}{
+		{
+			expect: ExpectOperatorAvailable,
+			name:   "[k8s][openshift] Expect operator to be available",
+		},
+		{
+			expect: CreateClusterAutoscaler,
```
enxebre: why is this excluded from [k8s]?

enxebre: this should be consistent with the Expect* convention

test/resources/machine-set.yaml
```diff
@@ -0,0 +1,26 @@
+---
```
enxebre: I can't see where this is used

enxebre:
looks great! I dropped some questions, can you also link t counterpart ci job?

test/resources/machine-set.yaml
frobware: I currently use this manually when standing up a kubemark cluster so can we keep it please.

test/e2e/main.go
ingvagabund: Since I am creating different ClusterAutoscaler resources (more constrained one) so the scaling down is quick.

test/resources/machine-set.yaml
ingvagabund: in the CI to create machine set

test/e2e/main.go
enxebre: what's exactly "I am creating", is there any code somewhere, in the jj may be? couldn't it be created here analogous to [openshift]?

test/resources/machine-set.yaml
enxebre: Can we move the CI code that instantiate this into here so upcoming changes can be done atomically in one repo?

ingvagabund: We can't. The test assumes a machine set already exists. The goal is to eventually run it over openshift as well. Thus, make it uniform.

enxebre: so before the actual validation runs, why can't we instantiate the machineSet here?

test/e2e/main.go
ingvagabund: Compared to the ClusterAutoscaler for openshift I need following additions to it:

```go
ca.Spec.ResourceLimits.Cores.Min = 1
ca.Spec.ResourceLimits.Memory.Min = 1
ca.Spec.ScaleDown = &autoscalingv1alpha1.ScaleDownConfig{
	Enabled:           true,
	DelayAfterAdd:     "10s",
	DelayAfterDelete:  "10s",
	DelayAfterFailure: "10s",
	UnneededTime:      "10s",
}
```

frobware: I also explicitly added this in https://github.com/openshift/kubernetes-autoscaler/pull/40/commits/3295fc36c105bf329c9ed90acf621482002585a4#diff-d73fd64bab7fd758991b5a15f551fd02R143

test/resources/machine-set.yaml
ingvagabund: My goal was to keep cao unaware of underlying machines or machinesets. Cao is responsible only for rendering manifests specified by ClusterAutoscaler and MachineAutoscaler resources. The less knowledge we expose, the better. Thus, the reason to have machineset created independently of running the tests. We need to distinguish between expectations of what always has to be present in the cluster and what the tests are responsible for creating (and deleting).

test/e2e/main.go
ingvagabund: I just copy pasted what was already in the job. I may change the name if it's needed though we need to be careful which expectations are valid. At this point it's just about naming though it would make more sense to do `ExpectClusterAutoscalerToBeCreated`. In general "`Expect<Object>ToBe<State>`" or "`Expect<Object>To<Operator>`".

enxebre: I'm confused here, is there anything stopping us from declaring this in code, exactly same we are doing for openshift and just set different param values

enxebre:
In a follow up I think we should:
- Move the instantiation of the clusterAutoscaler resource up to the `CreateClusterAutoscaler`
- Run `CreateClusterAutoscaler` for [k8s, OpenShift]
- Run `ExpectToScaleUpAndDown` for [k8s, OpenShift]
For reference JJ:
https://github.com/openshift/aos-cd-jobs/pull/1693/files#diff-fc4e56b6437666c4e7421a4224cbc443R38

Can we make sure we run minikube on 1.12? then I'm happy to approve

PR #50, opened 2019-02-18
Add expect for cluster operator status

enxebre:
- Introduce expect for cluster operator status 
- Fix dependency cluster operator status name, machine-api
- Better logging

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -1,6 +1,7 @@
 package operator
 
 import (
+	"github.com/golang/glog"
 	"time"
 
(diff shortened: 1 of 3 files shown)


enxebre:
cc @bison

ingvagabund:
Looks good. It's basically copy-paste of what we have in mao [1], right? Let's re-consider moving this test into common repo and vendoring it back here. Not a blocker for the PR.

[1] https://github.com/openshift/machine-api-operator/blob/master/test/e2e/operator_expectations.go#L73

/approve

ingvagabund:
`no kind is registered for the type v1.ClusterOperator`

PR #51, opened 2019-02-18
Rename labels from sigs.k8s.io to machine.openshift.io

spangenberg:
/cc bison


ingvagabund:
> I think tests will not pass until some of the other PRs land though.

`test/resources/ is used only by the kubemark CI so it should be fine

PR #52, opened 2019-02-22
Add OpenAPI schema validation

bison:
This adds OpenAPI schema validation via `kubebuilder` comments. The code generation switches back to using the operator-sdk (version v0.5.0 or newer) to enable this.


bison:
/hold cancel

Did some manual testing. Seems good to me.

install/0000_50_cluster-autoscaler-operator_01_clusterautoscaler.crd.yaml
```diff
@@ -1,8 +1,7 @@
-# TODO(bison): Add schema.
----
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
+  creationTimestamp: null
```
frobware: Why do we need this?

ingvagabund: Generated automatically

bison: The CRDs are generated now, by operator-sdk. It adds that bit. We don't need it, but it also doesn't hurt anything.

PR #53, opened 2019-02-25
Add configuration option for autoscaler logging verbosity

bison:
Adds configurable verbosity for autoscaler logging.

### DIFF
--- a/pkg/operator/config.go
+++ b/pkg/operator/config.go
@@ -43,6 +43,10 @@ const (
 	// DefaultClusterAutoscalerCloudProvider is the default name for
 	// the CloudProvider beeing used.
 	DefaultClusterAutoscalerCloudProvider = "openshift-machine-api"
+
+	// DefaultClusterAutoscalerVerbosity is the default logging
+	// verbosity level for ClusterAutoscaler deployments.
+	DefaultClusterAutoscalerVerbosity = 1
 )
 
 // Config represents the runtime configuration for the operator.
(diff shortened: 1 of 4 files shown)


pkg/operator/config.go
```diff
@@ -43,6 +43,10 @@ const (
 	// DefaultClusterAutoscalerCloudProvider is the default name for
 	// the CloudProvider beeing used.
 	DefaultClusterAutoscalerCloudProvider = "openshift-machine-api"
+
+	// DefaultClusterAutoscalerVerbosity is the default logging
+	// verbosity level for ClusterAutoscaler deployments.
+	DefaultClusterAutoscalerVerbosity = 2
```
ingvagabund: at least 3 please :)

pkg/operator/config.go
```diff
@@ -140,5 +149,15 @@ func ConfigFromEnvironment() *Config {
 		config.ClusterAutoscalerNamespace = caNamespace
 	}
 
+	if caVerbosity, ok := os.LookupEnv("CLUSTER_AUTOSCALER_VERBOSITY"); ok {
```
ingvagabund: +1

pkg/operator/config.go
frobware: I think the consensus on the channel was 1. so, LGTM.

bison: Done.

PR #56, opened 2019-02-26
Add tests and change StatusReporter.client to osconfig.Interface

michaelgugino:
Add status_test.go unit tests.  Also, change type of
StatusReporter.client to osconfig.Interface to allow for
utilizing fakeclient library.


enxebre:
/hold
@michaelgugino can you have a separate commit for vendoring, thanks?

pkg/operator/status_test.go
```diff
@@ -0,0 +1,61 @@
...
+					Type:   configv1.OperatorAvailable,
+					Status: osconfigv1.ConditionTrue,
+				},
+				{
+					Type:   configv1.OperatorFailing,
+					Status: osconfigv1.ConditionFalse,
+				},
+			},
+		},
+		{
```
ingvagabund: This item is identical to the previous one. Is that intentional?

michaelgugino: Good catch!  No, I don't know how that happend!

michaelgugino: @ingvagabund fixed.

michaelgugino:
@enxebre added separate commit for vendor.

PR #59, opened 2019-02-27
Override default CAO image

ingvagabund:
Currently specified image docker.io/openshift/origin-cluster-autoscaler-operator:v4.0 is not available.
Using the one in quay instead.

### DIFF
--- a/cluster-autoscaler-operator-patch.yaml
+++ b/cluster-autoscaler-operator-patch.yaml
@@ -9,6 +9,7 @@ spec:
     spec:
       containers:
       - name: cluster-autoscaler-operator
+        image: quay.io/openshift/origin-cluster-autoscaler-operator:v4.0
         env:
         - name: WATCH_NAMESPACE
           value: kube-system


bison:
/hold

Holding in case the tests pass. This must be changed in the image-references file as well or everything will break.

bison:
/hold cancel

Oops, this is in the patch. Sorry. Could we move this under `test/` or something?

ingvagabund:
> Oops, this is in the patch. Sorry. Could we move this under `test/` or something?

It's one way. Though, I would rather move `/install` directory under `/config` and then move both `kustomization.yaml` file and the patch under `/config`. @enxebre is there any limitation on when the `install` dir needs to be? It should not matter as long as it's properly install in the docker image, right?

bison:
There is no limitation as long as it goes into `/manifests` in the image. I don't think any other operator repo is putting these under `config/` though. They are all under `install` or `manifests`. We also can't put anything that's not a CVO manifest in that directory.

PR #60, opened 2019-02-27
Vendor ginkgo tests

ingvagabund:
- have `make test-e2e` run ginkgo test suite


Gopkg.toml
```diff
@@ -8,36 +8,35 @@ required = [
...
+  "github.com/openshift/cluster-api-actuator-pkg/pkg/e2e",
+  "github.com/onsi/ginkgo",
+  "github.com/onsi/gomega",
 ]
 
 [[override]]
   name = "k8s.io/code-generator"
-  # revision for tag "kubernetes-1.11.2"
-  revision = "6702109cc68eb6fe6350b83e14407c8d7309fd1a"
+  version = "kubernetes-1.12.5"
```
michaelgugino: I assume these tags are treated as immutable upstream.  Did you use a tool or a source-of-truth for picking this particular version?  I know we discussed replicating what origin was currently using.  Might be good to stick a comment at the top of the file or somewhere as to the how/why.

ingvagabund: > I assume these tags are treated as immutable upstream

I strongly hope they are. Assumption is you will always get the same revision.

> Did you use a tool or a source-of-truth for picking this particular version? I know we discussed replicating what origin was currently using. Might be good to stick a comment at the top of the file or somewhere as to the how/why.

There is a dashboard with versions of k8s dependencies of all components, where `v1.12.5` was the newest one. Otherwise, I don't think there is any rule which version should be picked.

ingvagabund: I am dropping any change to the k8s version from the PR though.

PR #61, opened 2019-02-27
Allow additional arguments to cluster-autoscaler

bison:
Allows additional arguments to the autoscaler for debugging purposes. This is not exposed in the CRD and setting this on the operator deployment will be overwritten by the CVO unless an override is explicitly added.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -383,7 +383,7 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:d562e86a00e1439011e117b30a4f5872fe4ed448044c4bceaae2d9bbd64b3738"
+  digest = "1:fda26b4b3cc8dc89575c57b4c141db0cf03051c65a7296e143683e1bd8aa9962"
   name = "github.com/openshift/cluster-api-actuator-pkg"
   packages = [
     "pkg/e2e",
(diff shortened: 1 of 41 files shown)


bison:
@enxebre @derekwaynecarr: What is your opinion on this? We want this for debugging. This allows adding arbitrary arguments to autoscaler that would normally be removed by the operator.

That is not desirable in the normal case of course, but you have to go out of your way to use this, e.g. by disabling the CVO or adding an override. Otherwise editing the edit to the operator deployment to add the environment variable would be reverted by the CVO. So I don't think it's really a big support burden.

frobware:
> @enxebre @derekwaynecarr: What is your opinion on this? We want this for debugging.

LGTM in that I have been using that to debug various autoscaler issues. But pending a real lgtm based on follow up from others ^^

ingvagabund:
> // This is not exposed in the CRD.  It is only configurable via
// environment variable, and in a normal OpenShift install the CVO
// will remove it if set manually.  It is only for development and
// debugging purposes.

Good enough for me

/approve

enxebre:
https://github.com/openshift/cluster-autoscaler-operator/pull/61/files#diff-614eabe934833da7d7ec9938dfb3f1faR96 thanks /lgtm

PR #62, opened 2019-02-28
Don't use an embedded ResourceRange in GPULimits

bison:
It looks like the CRD + OpenAPI generation in [controller-tools](https://github.com/kubernetes-sigs/controller-tools) is not handling embedded structs well. The min and max fields were missing for GPU limits in the ClusterAutoscaler resource.

This is an attempt at a quick fix by just adding explicit min and max fields to the type. I'm not sure this is the right thing to do though. It's kind of an API change, though I don't think anyone would notice at this point.


ingvagabund:
As long as the CRD is still fully generated, I have no objections.

Not sure how many consumers of the gpu resource limit we have, I would assume none.

frobware:
@bison Will the example need updating? This is where I ran into this issue.

bison:
@frobware, nope, it shouldn't. This should fix things so that the example works as is again.

PR #63, opened 2019-02-28
Add unit tests for Apply

michaelgugino:


pkg/operator/status_test.go
```diff
@@ -55,7 +55,41 @@ func TestCheckMachineAPI(t *testing.T) {
 			relatedObjects: []configv1.ObjectReference{},
 		}
 		res, err := r.CheckMachineAPI()
-		assert.Equal(t, res, tc.expectedBool, "case %v: return expected %v but didn't get it", i, tc.expectedBool)
-		assert.Equal(t, err, tc.expectedErr, "case %v: expected %v error but didn't get it, got: ", i, tc.expectedErr, err)
+		assert.Equal(t, tc.expectedBool, res, "case %v: return expected %v but didn't get it", i, tc.expectedBool)
```
michaelgugino: order should be 'expected, actual' for assert.Equal.

frobware: Did we lose an assertion here?

PR #65, opened 2019-02-28
Add status=progressing for version difference

michaelgugino:
Fixes: https://jira.coreos.com/browse/CLOUD-392

Depends-on: https://github.com/openshift/cluster-autoscaler-operator/pull/63

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/golang/glog"
 	configv1 "github.com/openshift/api/config/v1"
+	osconfigv1 "github.com/openshift/api/config/v1"
 	osconfig "github.com/openshift/client-go/config/clientset/versioned"
 	"github.com/openshift/cluster-autoscaler-operator/pkg/version"
 	cvorm "github.com/openshift/cluster-version-operator/lib/resourcemerge"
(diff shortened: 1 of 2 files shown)


pkg/operator/status.go
```diff
@@ -191,3 +246,11 @@ func (r *StatusReporter) CheckMachineAPI() (bool, error) {
 	glog.Infof("machine-api-operator not ready yet")
 	return false, nil
 }
+
+func printOperandVersions(desiredVersions []osconfigv1.OperandVersion) string {
```
frobware: nit: on `print`. it doesn't actually print, so maybe just `operandVersions()`?

pkg/operator/status_test.go
```diff
@@ -93,3 +93,62 @@ func TestApplyConditions(t *testing.T) {
...
+}
+
+func TestPrintOperandVersions(t *testing.T) {
+	desiredVersions := []osconfigv1.OperandVersion{
+		{
+			Name:    "operator",
+			Version: "2.0",
+		},
+	}
+	expectedString := "operator: 2.0"
```
frobware: Given that `[print]operandVersions()` joins the `desiredVersions` together does it make sense to expand the test cases here to cover that case?

pkg/operator/status.go
michaelgugino: This is copied from machine-api-operator.  I'd rather keep the naming the same for that reason.

pkg/operator/status_test.go
michaelgugino: Good idea.

smarterclayton:
Pretty sure this is broken - it hasn't passed a single time

/hold

ingvagabund:
Still looks good
/lgtm

PR #66, opened 2019-03-01
status: Only report available when rollout is complete at new version

smarterclayton:
Report the release version when all deployments are updated to the latest,
to prevent the CO reporting "good" when nothing has been upgraded yet.

We are moving to a new protocol for reporting readiness to the CVO.  See https://github.com/openshift/cluster-image-registry-operator/pull/219 for more context.

Also move the autoscaler to the default run level, it's not special.

@derekwaynecarr


install/04_deployment.yaml
```diff
@@ -26,6 +26,8 @@ spec:
         args:
         - -alsologtostderr
         env:
+        - name: RELEASE_NAMESPACE
```
derekwaynecarr: You mean version, not namespace.

michaelgugino:
/hold

@smarterclayton I will pick this up, I have another patch that needs to ship first and I'll reconcile.

install/04_deployment.yaml
smarterclayton: Yeah

michaelgugino:
Closing in favor of: https://github.com/openshift/cluster-autoscaler-operator/pull/67

PR #67, opened 2019-03-01
status: Only report available when rollout is complete at new version

michaelgugino:
Report the release version when all deployments are updated to the latest,
to prevent the CO reporting "good" when nothing has been upgraded yet.

We are moving to a new protocol for reporting readiness to the CVO. See [openshift/cluster-image-registry-operator#219](https://github.com/openshift/cluster-image-registry-operator/pull/219) for more context.

Also move the autoscaler to the default run level, it's not special.

Obsoletes: https://github.com/openshift/cluster-autoscaler-operator/pull/66
Depends-On: https://github.com/openshift/cluster-autoscaler-operator/pull/65

cc: @smarterclayton @derekwaynecarr


enxebre:
thanks, looking great! can you please update the description and point to previews PR as well https://github.com/openshift/cluster-autoscaler-operator/pull/66

pkg/operator/status.go
```diff
@@ -156,14 +194,22 @@ func (r *StatusReporter) Report(stopCh <-chan struct{}) error {
...
-		if ok {
-			r.Available(ReasonEmpty, "")
-			return true, nil
+		if !ok {
+			r.Fail(ReasonMissingDependency, "machine-api operator not ready")
+			return false, nil
+		}
+		ok, err = check.AvailableAndUpdated()
+		if err != nil {
+			r.Fail(ReasonCheckAutoscaler, fmt.Sprintf("error checking autoscaler operator status %v", err))
```
smarterclayton: I missed a colon here `operator status: %v"`

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -243,7 +276,8 @@ func (r *Reconciler) AutoscalerDeployment(ca *autoscalingv1alpha1.ClusterAutosca
 	}
 
 	annotations := map[string]string{
-		criticalPod: "",
+		criticalPod:                    "",
+		"release.openshift.io/version": r.config.ReleaseVersion,
```
smarterclayton: Ah, this is why.  The annotation should be on the Deployment, not the DeploymentSpec.metadata (I assumed it was setting on both).

That's good, we validated that the CVO waits!

michaelgugino: Updated.

ingvagabund:
@michaelgugino can you update `kustomization.yaml` file as well?

pkg/operator/config.go
```diff
@@ -107,6 +109,7 @@ func NewConfig() *Config {
 		ClusterAutoscalerReplicas:      DefaultClusterAutoscalerReplicas,
 		ClusterAutoscalerCloudProvider: DefaultClusterAutoscalerCloudProvider,
 		ClusterAutoscalerVerbosity:     DefaultClusterAutoscalerVerbosity,
+		ReleaseVersion:                 DefaultReleaseVersion,
```
smarterclayton: We should never report this value - so if you want this sort of marker value you need to fail before you set it.  It would be illegal to set a non-semantic version into the operator.

PR #69, opened 2019-03-07
Operator: move report logic to ApplyStatus function

michaelgugino:
This commit refactors business logic of Report() to
it's own function, ApplyStatus().  This assists
in unit testing.

This test also implements behavior validation for
critical status reporting logic.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -23,13 +23,13 @@ const (
 	ReasonCheckAutoscaler   = "UnableToCheckAutoscalers"
 )
 
-// ApplyStatusInterface is used to enable unit testing and does not
+// applyStatusInterface is used to enable unit testing and does not
 // implement all methods of StatusReporter.
-type ApplyStatusInterface interface {
-	CheckMachineAPI() (bool, error)
-	Fail(string, string) error
-	Progressing() error
-	Available(string, string) error
+type applyStatusInterface interface {
+	checkMachineAPI() (bool, error)
+	fail(string, string) error
+	progressing() error
+	available(string, string) error
 }
 
 // StatusReporter reports the status of the operator to the OpenShift
@@ -114,9 +114,9 @@ func (r *StatusReporter) ApplyConditions(conds []configv1.ClusterOperatorStatusC
 	return err
 }
 
-// Available reports the operator as available, not progressing, and
+// available reports the operator as available, not progressing, and
 // not failing -- optionally setting a reason and message.
-func (r *StatusReporter) Available(reason, message string) error {
+func (r *StatusReporter) available(reason, message string) error {
 	conditions := []configv1.ClusterOperatorStatusCondition{
 		{
 			Type:    configv1.OperatorAvailable,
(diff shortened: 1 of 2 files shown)


pkg/operator/status_test.go
```diff
@@ -60,6 +61,190 @@ func TestCheckMachineAPI(t *testing.T) {
...
+	} else {
+		return false, nil
+	}
+}
+
+func (r *MockStatusReporter) CheckMachineAPI() (bool, error) {
+	if r.isCheckMachineAPIFail {
+		return false, errors.New("returning failure")
+	}
+	if r.isCheckMachineAPI {
```
frobware: Collapse to `return r.isCheckMachineAPI, nil`.

pkg/operator/status.go
```diff
@@ -23,6 +23,15 @@ const (
 	ReasonCheckAutoscaler   = "UnableToCheckAutoscalers"
 )
 
+// StatusReporterInterface is used to enable unit testing and does not
+// implement all methods of StatusReporter.
+type StatusReporterInterface interface {
```
frobware: I find the name confusing as we already have `StatusReporter`. Can we differentiate with another name?

pkg/operator/status_test.go
```diff
@@ -60,6 +61,190 @@ func TestCheckMachineAPI(t *testing.T) {
...
+type MockCheck struct {
+	isFail                bool
+	isAvailableAndUpdated bool
+}
+
+func (c *MockCheck) AvailableAndUpdated() (bool, error) {
+	if c.isFail {
+		return false, errors.New("returning isFail")
+	}
+	if c.isAvailableAndUpdated {
```
frobware: Collapse to `return c.isAvailableAndUpdated, nil`.

pkg/operator/status_test.go
```diff
@@ -60,6 +61,190 @@ func TestCheckMachineAPI(t *testing.T) {
...
+	} else {
+		return false, nil
+	}
+}
+
+func (r *MockStatusReporter) Available(reason, message string) error {
+	r.availableCalled = true
+	r.availableReason = reason
+	r.availableMessage = message
+	if r.isAvailable {
```
frobware: nit: Deal with the error cases first?

```go
if !r.isAvailable ...
```

pkg/operator/status.go
ingvagabund: Can you make the interface private? It's fine to use it in unit tests. Though, nothing to be consumed outside of `pkg/operator` package.

pkg/operator/status.go
```diff
@@ -23,6 +23,15 @@ const (
...
+	CheckMachineAPI() (bool, error)
+	Fail(string, string) error
+	Progressing() error
+	Available(string, string) error
+}
```
ingvagabund: I see all of the listed methods to be used solely in `pkg/operator` package. Can we make all of them private?

pkg/operator/status.go
```diff
@@ -180,6 +190,34 @@ func (r *StatusReporter) Progressing(reason, message string) error {
 	return r.ApplyConditions(conditions, false)
 }
 
+// Interface required here for test coverage of critical code.
+func ApplyStatus(r StatusReporterInterface, check AvailableChecker) (bool, error) {
```
ingvagabund: Could be a private function as well.

pkg/operator/status_test.go
```diff
@@ -60,6 +61,182 @@ func TestCheckMachineAPI(t *testing.T) {
...
+	return r.isCheckMachineAPI, nil
+}
+
+func (r *MockStatusReporter) Available(reason, message string) error {
+	r.availableCalled = true
+	r.availableReason = reason
+	r.availableMessage = message
+	if !r.isAvailable {
+		return errors.New("returning failure")
+	} else {
```
frobware: Drop the else and just `return nil`.

PR #70, opened 2019-03-08
Refactor CAS_controller available check and tests

michaelgugino:
This commit refactors clusterautoscaler_controller.AvailableAndUpdated
in order to support better testing.

This commit also implements unit tests for critical section of
logic for determining if operand (clusterautoscaler) is ready
in order to properly report status=progressing/available.

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
@@ -57,6 +57,13 @@ type Config struct {
 	Verbosity int
 }
 
+// deploymentUpdatedInterface is used to enable unit testing and does not
+// implement all methods of Reconciler.
+type deploymentUpdatedInterface interface {
+	GetAutoscaler(ca *autoscalingv1alpha1.ClusterAutoscaler) (*appsv1.Deployment, error)
+	isDeploymentControllerCurrent(dep *appsv1.Deployment) (bool, error)
+}
+
 var _ reconcile.Reconciler = &Reconciler{}
 
 // Reconciler reconciles a ClusterAutoscaler object
@@ -171,6 +178,16 @@ func (r *Reconciler) AvailableAndUpdated() (bool, error) {
 		}
 		return false, err
 	}
+	return isDeploymentUpdated(r, ca, r.config)
+}
+
+func (r *Reconciler) isDeploymentControllerCurrent(dep *appsv1.Deployment) (bool, error) {
+	return (dep.Status.ObservedGeneration >= dep.Generation && dep.Status.UpdatedReplicas == dep.Status.Replicas && dep.Status.AvailableReplicas > 0), nil
+}
+
+// deploymentUpdatedInterface required here for test coverage of critical code.
+// In actual operation, we are passing in Reconciler.
+func isDeploymentUpdated(r deploymentUpdatedInterface, ca *autoscalingv1alpha1.ClusterAutoscaler, rconfig *Config) (bool, error) {
 	dep, err := r.GetAutoscaler(ca)
 	if err != nil {
 		if errors.IsNotFound(err) {
@@ -179,15 +196,11 @@ func (r *Reconciler) AvailableAndUpdated() (bool, error) {
 		}
 		return false, err
 	}
-	if dep.ObjectMeta.Annotations["release.openshift.io/version"] != r.config.ReleaseVersion {
+	if dep.ObjectMeta.Annotations["release.openshift.io/version"] != rconfig.ReleaseVersion {
 		// still haven't synced the release version
 		return false, nil
 	}
-	if dep.Status.ObservedGeneration < dep.Generation || dep.Status.UpdatedReplicas != dep.Status.Replicas || dep.Status.AvailableReplicas == 0 {
-		// deployment hasn't rolled out a new controller or we still have the old version hanging around
-		return false, nil
-	}
-	return true, nil
+	return r.isDeploymentControllerCurrent(dep)
 }
 
 // SetConfig sets the given config on the reconciler.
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterautoscaler/clusterautoscaler_test.go
```diff
@@ -125,3 +132,120 @@ func TestAutoscalerArgs(t *testing.T) {
...
+}
+
+func (r *MockReconciler) GetAutoscaler(_ *autoscalingv1alpha1.ClusterAutoscaler) (*appsv1.Deployment, error) {
+	dep := &appsv1.Deployment{}
+
+	if !r.getAutoscalerOk {
+		return nil, r.gASerrType
+	}
+	var m map[string]string
+	m = make(map[string]string)
```
frobware: Just make this `m := map[string]string{}`.

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
```diff
@@ -125,3 +132,120 @@ func TestAutoscalerArgs(t *testing.T) {
...
+	dep.ObjectMeta.Annotations = m
```
frobware: Why can't this just be `dep.ObjectMeta.Annotations = map[string]string{}`?

And/or:

```go
if dep.ObjectMeta.Annotations == nil {
    dep.ObjectMeta.Annotations = map[string]string{}
}
```

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
```diff
@@ -125,3 +132,120 @@ func TestAutoscalerArgs(t *testing.T) {
...
+	return r.isDCCResult, nil
+}
+
+// This test ensures we can actually get an autoscaler with fakeclient/client.
+// fakeclient.NewFakeClientWithScheme will os.Exit(1) with invalid scheme.
+func TestCanGetca(t *testing.T) {
+	tscheme := runtime.NewScheme()
+	apis.AddToScheme(tscheme)
+	ca := NewClusterAutoscaler()
+	_ = fakeclient.NewFakeClientWithScheme(tscheme, ca)
```
frobware: Drop the temporary; `_ = fakeclient.NewFakeClientWithScheme(tscheme, NewClusterAutoscaler())`.

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
```diff
@@ -125,3 +132,120 @@ func TestAutoscalerArgs(t *testing.T) {
 		}
 	}
 }
+
+type MockReconciler struct {
+	getAutoscalerOk bool
+	gASerrType      error
+	configVersion   string
+	isDCCResult     bool
+	calledIsDCC     bool
```
frobware: What is `DCC`? Is this `isCurrent`?

michaelgugino: There's a function named "isDeploymentControllerCurrent"

So this variable is calledIsDCC, eg, did you call isDeploymentControllerCurrent?

frobware: I read the two variables `isDCCResult` and `calledisDCC` and couldn't really grok what they were.

For `calledIsDCC` how about `isDCCFuncCalled`? Then it code it (possibly) reads as:

```go
if `isDCCFuncCalled` ...
```

Not sure - asking / raising as it was not immediately obvious what the two variables were doing.

frobware: Or even use `has` or `is`. /shrug

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
michaelgugino: I can collapse, but maps must be initialized.  Also, this should only be called once for each test, so if statement should be unnecessary.

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
enxebre: using the whole name instead of acronym  would be more readable

michaelgugino: The function it's called starts with 'is' to indicated a true/false scenario for the actual code.

isDeploymentControllerCurrent?  Yes, it is (true), not it's not (false).

isDCCResult = our contrived function's return value.  I agree, the naming is not the best, but in the production code I think it reads cleanly; isDCCResult is only used here in testing.

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -57,6 +57,13 @@ type Config struct {
 	Verbosity int
 }
 
+// deploymentUpdatedInterface is used to enable unit testing and does not
+// implement all methods of Reconciler.
+type deploymentUpdatedInterface interface {
```
enxebre: Not totally sure about the value of the interface here. Couldn't  we create a fake runtime client and test against multiple dep autoscaler specs and also use the value ["release.openshift.io/version"] coming from there so we use it to validate cases? cc @bison

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
frobware: > isDCCResult = our contrived function's return value. I agree, the naming is not the best, but in the production code I think it reads cleanly; isDCCResult is only used here in testing.

I wouldn't make a distinction between "production" and test code. Both are production, IMO.

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
bison: Agreed. Don't want to be a pain here, but I think it's important we keep the strategy aligned with our other operators and the other OpenShift operators. That's generally testing inputs and outputs via the fake k8s client rather than overriding individual methods for different test cases.

michaelgugino:
Closing in favor of: https://github.com/openshift/cluster-autoscaler-operator/pull/71

PR #71, opened 2019-03-13
ca_controller: implement unit tests

michaelgugino:
Add unit tests for AvailableAndUpdated() to ensure
cao status=progressing/available logic is correct.


bison:
We should consider adding a `-v` to the test target in the Makefile. The output is pretty minimal right now.

michaelgugino:
> We should consider adding a `-v` to the test target in the Makefile. The output is pretty minimal right now.

Output gets nice and verbose if a test fails :)

Also, I usually run with go test ./pkg/... directly, thus I just add -test.v directly.

PR #73, opened 2019-03-14
ca_controller: Remove duplicate api call in updateAutoscaler

michaelgugino:
This commit removes extra call to api to get the existing
deployment.  This deployment is already retrieved in
Reconcile() method.

This commit also changes UpdateAutoscaler to private
updateAutoscaler.

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
@@ -137,8 +137,8 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		glog.Errorf("Error reading ClusterAutoscaler: %v", err)
 		return reconcile.Result{}, err
 	}
-
-	_, err = r.GetAutoscaler(ca)
+	var dep *appsv1.Deployment
+	dep, err = r.GetAutoscaler(ca)
 	if err != nil && !errors.IsNotFound(err) {
 		glog.Errorf("Error getting cluster-autoscaler deployment: %v", err)
 		return reconcile.Result{}, err
@@ -153,7 +153,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{}, nil
 	}
 
-	if err := r.UpdateAutoscaler(ca); err != nil {
+	if err := r.updateAutoscaler(ca, dep); err != nil {
 		glog.Errorf("Error updating cluster-autoscaler deployment: %v", err)
 		return reconcile.Result{}, err
 	}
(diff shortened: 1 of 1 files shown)


pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -153,7 +153,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{}, nil
 	}
 
-	if err := r.UpdateAutoscaler(ca); err != nil {
+	if err := r.updateAutoscaler(ca, dep); err != nil {
```
ingvagabund: Both `r.GetAutoscaler` are invoked pretty close to each other so the chance of changing the deployment between the calls is quite small. And, if that happens and `client.Update` complains about `the object has been modified; please apply your changes to the latest version and try again`, the deployment gets reconciled in the next loop.

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -225,13 +225,9 @@ func (r *Reconciler) CreateAutoscaler(ca *autoscalingv1alpha1.ClusterAutoscaler)
...
 
-// UpdateAutoscaler will retrieve the deployment for the given ClusterAutoscaler
+// updateAutoscaler compares the deployment for the given ClusterAutoscaler
 // custom resource instance and update it to match the expected spec if needed.
-func (r *Reconciler) UpdateAutoscaler(ca *autoscalingv1alpha1.ClusterAutoscaler) error {
-	existingDeployment, err := r.GetAutoscaler(ca)
-	if err != nil {
-		return err
-	}
+func (r *Reconciler) updateAutoscaler(ca *autoscalingv1alpha1.ClusterAutoscaler, existingDeployment *appsv1.Deployment) error {
```
ingvagabund: +1 for making it public

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
michaelgugino: Actually, they are invoked after checks of the err with no other operations in between.  If line 141 returns a deployment, the very next action (other than two boolean statements which resolve false) is to call GetAutoscaler again!

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
michaelgugino: *private.

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
bison: I get the reasoning here, but I don't think this is performance critical and we have caching. This makes calling the update harder and more error prone if it were to be called elsewhere. I think the original focus was on the API being friendly.

bison: So, if we're going to make `updateAutoscaler()` private and require passing in the deployment to compare, my vote would probably be to just inline it in `Reconcile()` instead.

michaelgugino: @bison I disagree.  I'm not sure what context we would potentially call updateAutoscaler() if we didn't already have a deployment that we needed to update.

I don't agree with inlining, I like it as-is for now.

PR #75, opened 2019-03-15
Fix bug with filtering supported MachineAutoscaler GVKs

bison:
Fixes an issue with inadvertently modifying the SupportedGVKs in-place.


ingvagabund:
/approve

Useful refactoring

PR #77, opened 2019-03-21
Add additional logging

michaelgugino:
This commit adds some extra logging to ensure we capture
information during availability checks.

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
@@ -134,6 +134,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 			// reconcile request.  Owned objects are automatically
 			// garbage collected. For additional cleanup logic use
 			// finalizers.  Return and don't requeue.
+			glog.Infof("ClusterAutoscaler %s not found, will not reconcile", request.Name)
 			return reconcile.Result{}, nil
 		}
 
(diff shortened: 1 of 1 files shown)


ingvagabund:
+1 for also reporting a situation when nothing is happening

/approve

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -134,6 +134,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 			// reconcile request.  Owned objects are automatically
 			// garbage collected. For additional cleanup logic use
 			// finalizers.  Return and don't requeue.
+			glog.Info("Request not found, don't requeue")
```
bison: I'm not sure this log is that useful in practice. If anything, I'd make it `V(2)` or higher, but at a minimum it should have the `request.NamespacedName` attached. Something like:
```
glog.Infof("%s: Not found, will not reconcile", request.NamespacedName)
```

michaelgugino: No need, IMO.  We only operate on one namespace, and we capture the name in the "Reconciling Clusterautoscaler" log line above.

Since we notify via normal Info logging that we're attempting to reconcile something, we might as well log what the actual outcome was.  This is in an effort to better aid troubleshooting of a current BZ.

bison: Given that this resource is cluster scoped, we don't really need the NamespacedName, but if we're going to log this we absolutely should indicate that what was not found was a ClusterAutoscaler resource with a given name, so maybe instead:
```
glog.Infof("ClusterAutoscaler %s not found, will not reconcile")
```
We can't count on the log lines appearing directly adjacent to each other.

PR #79, opened 2019-03-27
Make StatusReporter a Runnable added to the manager

bison:
This refactors the StatusReporter to satisfy the Runnable interface
and adds it to the manager when creating a new Operator instance.
This prevents it from starting before clients are set up, and from
reporting status before leader-election has succeeded.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1689146

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/golang/glog"
 	autoscalingv1alpha1 "github.com/openshift/cluster-autoscaler-operator/pkg/apis/autoscaling/v1alpha1"
+	"github.com/openshift/cluster-autoscaler-operator/pkg/util"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -27,7 +28,6 @@ import (
 
 const (
 	controllerName      = "cluster-autoscaler-controller"
-	criticalPod         = "scheduler.alpha.kubernetes.io/critical-pod"
 	caServiceAccount    = "cluster-autoscaler"
 	caPriorityClassName = "system-cluster-critical"
 )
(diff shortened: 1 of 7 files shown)


bison:
Still WIP. Need to:
1. Actually update the release annotation in the controller.
2. Fix the tests.

pkg/operator/status.go
```diff
@@ -66,182 +91,225 @@ func (r *StatusReporter) GetOrCreateClusterOperator() (*configv1.ClusterOperator
...
+
+	// Block until the stop channel is closed.
+	<-stopCh
+
+	return err
+}
+
+// ReportStatus checks the status of each dependency and operand and reports the
+// appropriate status via the operator's ClusterOperator object.
+func (r *StatusReporter) ReportStatus() (bool, error) {
```
michaelgugino: This is a good place for passing in StatusReporter as an interface instead of using a method.  This would make mocking up the logic of this function easier, and it would decouple the implementation from the spec.

bison:
> There's a slight regression in test coverage. ReportStatus should have coverage similar to it's predecessor.

I think the constituent components were tested, but you're right, probably warrants its own set of tests. Done.

pkg/operator/status.go
bison: I get what you're saying, but I just don't think that necessary here. This is already easy to test by just mocking up the state of the things it checks like we do in the rest of the tests. I think that's a more realistic test as opposed to passing in a mock implementation.

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -290,6 +263,21 @@ func (r *Reconciler) AutoscalerName(ca *autoscalingv1alpha1.ClusterAutoscaler) t
 	}
 }
 
+// UpdateAnnotations updates the annoations on the given object to the values
```
enxebre: nit s/annoations/annotations

enxebre: would be nice a unit for this

pkg/operator/status.go
```diff
@@ -66,182 +92,225 @@ func (r *StatusReporter) GetOrCreateClusterOperator() (*configv1.ClusterOperator
...
-		status.Versions = []configv1.OperandVersion{
-			{
-				Name:    "operator",
-				Version: r.releaseVersion,
-			},
+	// Update LastTransitionTime for all conditions if necessary.
+	for i := range status.Conditions {
+		condType := status.Conditions[i].Type
+		timestamp := metav1.NewTime(time.Now())
+
```
enxebre: we should standardise on https://github.com/openshift/library-go/blob/master/pkg/config/clusteroperator/v1helpers/status.go

pkg/operator/status.go
```diff
@@ -66,182 +92,225 @@ func (r *StatusReporter) GetOrCreateClusterOperator() (*configv1.ClusterOperator
...
+}
+
+// ReportStatus checks the status of each dependency and operand and reports the
+// appropriate status via the operator's ClusterOperator object.
+func (r *StatusReporter) ReportStatus() (bool, error) {
+	// Check that the machine-api-operator is reporting available.
+	ok, err := r.CheckMachineAPI()
 	if err != nil {
-		r.fail(ReasonCheckAutoscaler, fmt.Sprintf("error checking autoscaler operator status: %v", err))
+		msg := fmt.Sprintf("error checking machine-api status: %v", err)
```
enxebre: nit: we just could inline msg here

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -183,38 +183,6 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
-			// waiting for deployment to be created
-			return false, nil
-		}
-		return false, err
-	}
-	if dep.ObjectMeta.Annotations["release.openshift.io/version"] != r.config.ReleaseVersion {
-		// still haven't synced the release version
-		return false, nil
-	}
-	if dep.Status.ObservedGeneration < dep.Generation || dep.Status.UpdatedReplicas != dep.Status.Replicas || dep.Status.AvailableReplicas == 0 {
```
enxebre: for the record, this is now tracked by `util.DeploymentUpdated`

PR #80, opened 2019-04-01
Record ClusterAutoscaler events in correct namespace

bison:
Because ClusterAutoscaler objects are cluster scoped, events related
to these objects were being recorded in the default namesapce.  This
causes the events to be recorded in the namespace that the deployment
is created in.

This also adds some unit tests for the `UpdateAnnotations` method.

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
@@ -269,7 +269,7 @@ func (r *Reconciler) AutoscalerName(ca *autoscalingv1alpha1.ClusterAutoscaler) t
 	}
 }
 
-// UpdateAnnotations updates the annoations on the given object to the values
+// UpdateAnnotations updates the annotations on the given object to the values
 // currently expected by the controller.
 func (r *Reconciler) UpdateAnnotations(obj metav1.Object) {
 	annotations := obj.GetAnnotations()
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/tools/reference"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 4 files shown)


pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -143,10 +144,15 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		return reconcile.Result{}, err
 	}
 
+	// caRef is a reference to the ClusterAutoscaler object, but with the
+	// namespace for cluster-autoscaler deployments set.  This keeps events
+	// generated for these cluster scoped objects out of the default namespace.
+	caRef := r.objectReference(ca)
```
frobware: Can `caRef` ever be `nil`?

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -363,3 +369,21 @@ func (r *Reconciler) AutoscalerPodSpec(ca *autoscalingv1alpha1.ClusterAutoscaler
...
+
+// objectReference returns a reference to the given object, but will set the
+// configured deployment namesapce if no namespace was previously set.  This is
+// useful for referencing cluster scoped objects in events without the events
+// being created in the default namespace.
+func (r *Reconciler) objectReference(obj runtime.Object) *corev1.ObjectReference {
+	ref, err := reference.GetReference(r.scheme, obj)
+	if err != nil {
+		glog.Errorf("Error creating object reference: %v", err)
+		return nil
```
frobware: See: https://github.com/openshift/cluster-autoscaler-operator/pull/80#discussion_r271179412

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
bison: In practice it won't here, but yes it can. It will just cause subsequent event recording to fail and log a warning in this reconciliation. It won't affect anything else.

ingvagabund: How can it be CAO expects CA to be under `request.NamespacedName`. Yet, when you get the CA object from the namespace, the namespace is set to `""`?

bison: Exactly, so if you pass that ClusterAutoscaler object with an empty namespace to an even recorder, it places the event (which are namespaced objects) into the default namespace. This avoids that by creating an ObjectReference  pointing to the ClusterAutoscaler but explicitly adding a namespace. That reference is only used by the event recorder.

The alternative is to give this operator permissions to manage events in the default namespace, and to not have the events show up if you list all events in the `openshift-machine-api` namespace.

bison: Sorry, maybe I misunderstood. The namespace field of the `NamespacedName` here is *always* empty because ClusterAutoscaler resources are cluster-scoped. They don't have an associated namespace, just like Node or ClusterOperator objects for example.

PR #82, opened 2019-04-02
Move from glog to klog

bison:
Moving from glog to klog to align with related projects.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -85,3 +85,7 @@ required = [
 [[constraint]]
   branch = "master"
   name = "github.com/golang/glog"
+
+[[constraint]]
+  name = "k8s.io/klog"
+  version = "0.2.0"
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -4,31 +4,44 @@ import (
 	"flag"
 	"runtime"
 
-	"github.com/golang/glog"
 	"github.com/openshift/cluster-autoscaler-operator/pkg/operator"
 	"github.com/openshift/cluster-autoscaler-operator/pkg/version"
-	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"k8s.io/klog"
 )
 
 func printVersion() {
-	glog.Infof("Go Version: %s", runtime.Version())
-	glog.Infof("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH)
-	glog.Infof("Version: %s", version.String)
+	klog.Infof("Go Version: %s", runtime.Version())
+	klog.Infof("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH)
+	klog.Infof("Version: %s", version.String)
 }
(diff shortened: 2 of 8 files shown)


Gopkg.toml
```diff
@@ -85,3 +85,7 @@ required = [
 [[constraint]]
   branch = "master"
   name = "github.com/golang/glog"
```
ingvagabund: Does `klog` replaces `glog` completely? If so, this can be removed.

cmd/manager/main.go
```diff
@@ -4,31 +4,44 @@ import (
...
+
+	// Sync the glog and klog flags.
+	flag.CommandLine.VisitAll(func(f1 *flag.Flag) {
+		f2 := klogFlags.Lookup(f1.Name)
+		if f2 != nil {
+			value := f1.Value.String()
+			f2.Value.Set(value)
+		}
+	})
+
```
ingvagabund: Should be safe to just use:
```go
flag.Set("logtostderr", "true")
klog.InitFlags(nil)
flag.Parse()
```

bison: This is from the example in the klog repo and how things like the cluster-api tools initialize things.

Gopkg.toml
bison: It's actually still used in some of the tests. Was going to fix that in another PR.

ingvagabund:
Can you remove `test/e2e` directory as well (e.g in a separate PR)? We no longer use that code and it's still importing `github.com/golang/glog`.

bison:
> Can you remove test/e2e directory as well (e.g in a separate PR)? We no longer use that code and it's still importing github.com/golang/glog.

Yep. Planned on a follow up PR for that.

cmd/manager/main.go
ingvagabund: No objection here, just noticed we no longer have to use that hack of iterating through the flags. So, it's worth simplifying it and testing if all the flags are still available and properly propagated to vendored glog.

test/e2e/operator_expectations.go
```diff
@@ -13,7 +14,7 @@ import (
 	cvorm "github.com/openshift/cluster-version-operator/lib/resourcemerge"
 	kappsapi "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
michaelgugino: maybe corev1 here?  Did you mean to add this change, seems a little out of place for this PR.

ingvagabund: Automatically generated

enxebre:
please let make sure's there's a follow up for getting rid of glog completely from the vendor and drop the VisitAll loop
/lgtm

test/e2e/operator_expectations.go
bison: Yeah, this is from running `goimports`.

PR #93, opened 2019-04-24
Add support for status=Degraded

michaelgugino:
This commit adds support for status=Degraded and deprecates
status=Failed for MAO availability check.

Depends-on: https://github.com/openshift/cluster-autoscaler-operator/pull/92


pkg/operator/status.go
```diff
@@ -328,7 +328,8 @@ func (r *StatusReporter) CheckMachineAPI() (bool, error) {
 	conds := mao.Status.Conditions
 
 	if cvorm.IsOperatorStatusConditionTrue(conds, configv1.OperatorAvailable) &&
-		cvorm.IsOperatorStatusConditionFalse(conds, configv1.OperatorFailing) {
+		(cvorm.IsOperatorStatusConditionFalse(conds, configv1.OperatorFailing) ||
```
bison: Is there any circumstance (even if it's not the ordinary) where `Failing` would be present and false, but `Degraded` would be true?

michaelgugino: @bison failing is being removed and replaced by degraded.  Unfortunately, we have to straddle this change like this.

bison: Yeah, understood. Just saying that if, for instance, a "left over" `Failing` condition existed that was set to false, we would never check the status of the `Degraded` condition.

michaelgugino: @bison I'm planning on removing this so we can get everything up to date, so we shouldn't be shipping this hopefully.  Just a CI band-aide.

PR #94, opened 2019-04-26
Add support for admission webhooks

bison:
This adds support for running validating / mutating admission
webhooks.  The configuration for the webhooks is expected to be
applied by the cluster-version-operator, and TLS assets are generated
by the service-ca-operator.

The service-ca-operator does not yet support injecting the CA bundle
into webhook configurations, so for now the operator reads the CA at
start-up and patches the webhook configuration resource with the CA.


install/07_webhooks.yaml
```diff
@@ -0,0 +1,28 @@
...
+  rules:
+  - apiGroups:
+    - autoscaling.openshift.io
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - clusterautoscalers
```
enxebre: I assume we'll have another one for machineautoscalers, right?

pkg/operator/operator.go
```diff
@@ -40,16 +52,24 @@ func New(cfg *Config) (*Operator, error) {
...
 	}
 
+	// Setup our controllers and add them to the manager.
 	if err := operator.AddControllers(); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to add controllers: %v", err)
+	}
+
+	// Setup admission webhooks and add them to the manager.
+	if cfg.WebhooksEnabled {
```
enxebre: the manifests are still created so what even when this is disable?

install/07_webhooks.yaml
bison: Yeah, eventually.

pkg/operator/operator.go
bison: You might still run the operator locally and not want the webhooks -- that's just one example. Also, the configuration is set to ignore failure, so even if it exists, you can choose not to enable the webhook.

install/03_rbac.yaml
```diff
@@ -4,6 +4,16 @@ kind: ClusterRole
...
   name: cluster-autoscaler-operator
 rules:
+- apiGroups:
+  - admissionregistration.k8s.io
+  resources:
+  - validatingwebhookconfigurations
+  - mutatingwebhookconfigurations
+  verbs:
+  - get
+  - patch
```
enxebre: will the service-ca-operator patch or update here?

bison: I'm not sure what the service-ca-operator does when it updates the `caBundle`. In the interim solution where the cluster-autoscaler-operator itself copies the CA from a ConfigMap to the webhook config, we use a PATCH, so I've added these permissions.

install/04_deployment.yaml
```diff
@@ -71,9 +71,9 @@ spec:
           secretName: cluster-autoscaler-operator-cert
           items:
           - key: tls.crt
-            path: cert.pem
+            path: tls.crt
           - key: tls.key
-            path: key.pem
```
enxebre: does this belong to this commit?

bison: Yes, these file names are actually hard coded in controller-runtime and they changed to match the keys in Kubernetes TLS secrets recently.

bison:
Closing this in favor of splitting it up into two phases: First upgrade controller-runtime to the tls-reload branch, then add webhook support.

PR #99, opened 2019-05-06
Gen crd

ingvagabund:
To provide richer output to `oc explain <CRD>` command. E.g.
```sh
$ oc explain clusterautoscaler.spec.scaleDownKIND:     ClusterAutoscaler
VERSION:  autoscaling.openshift.io/v1

RESOURCE: scaleDown <Object>

DESCRIPTION:
     Configuration of scale down operation

FIELDS:
   delayAfterAdd	<string>
     How long after scale up that scale down evaluation resumes

   delayAfterDelete	<string>
     How long after node deletion that scale down evaluation resumes, defaults
     to scan-interval

   delayAfterFailure	<string>
     How long after scale down failure that scale down evaluation resumes

   enabled	<boolean> -required-
     Should CA scale down the cluster

   unneededTime	<string>
     How long a node should be unneeded before it is eligible for scale down
```

```sh
$ kubectl explain machineautoscaler
KIND:     MachineAutoscaler
VERSION:  autoscaling.openshift.io/v1beta1

DESCRIPTION:

FIELDS:
   apiVersion	<string>
     APIVersion defines the versioned schema of this representation of an
     object. Servers should convert recognized schemas to the latest internal
     value, and may reject unrecognized values. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#resources

   kind	<string>
     Kind is a string value representing the REST resource this object
     represents. Servers may infer this from the endpoint the client submits
     requests to. Cannot be updated. In CamelCase. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds

   metadata	<Object>
     Standard object's metadata. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata

   spec	<Object>
     Specification of constraints of a scalable resource

   status	<Object>
     Most recently observed status of a scalable resource
```

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1705748


hack/gen-crd.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -eu
+
+echo "Building controller-gen tool..."
+go build -o bin/controller-gen github.com/openshift/cluster-autoscaler-operator/vendor/sigs.k8s.io/controller-tools/cmd/controller-gen
+
+bin/controller-gen crd --domain openshift.io
+
+#echo "Coping generated CRDs"
```
vikaschoudhary16: either uncomment or remove please.

Gopkg.toml
```diff
@@ -15,6 +15,7 @@ required = [
   "github.com/openshift/api/config/v1",
   "github.com/onsi/ginkgo",
   "github.com/onsi/gomega",
+  "sigs.k8s.io/controller-tools/cmd/controller-gen", # for crd/rbac generation
```
enxebre: couldn't we `go get` here or similar so no need to bring +29,122 lines?

ingvagabund: That's the way it is. `go get` would not help if we want to keep all the deps in `Gopkg.toml`.

enxebre: which deps? this is not used in code right? We need a `controller-gen` binary to leverage automation for the case where we want to regenerate CRDs. Why can't we fetch it instead of building it from source code?

pkg/apis/autoscaling/v1beta1/machineautoscaler_types.go
```diff
@@ -10,15 +10,21 @@ func init() {
 
 // MachineAutoscalerSpec defines the desired state of MachineAutoscaler
 type MachineAutoscalerSpec struct {
+	// MinReplicas constrains the minimal number of replicas
```
vikaschoudhary16: s/minimal number of replicas to be present all the time in the cluster during scale down?

pkg/apis/autoscaling/v1beta1/machineautoscaler_types.go
```diff
@@ -10,15 +10,21 @@ func init() {
...
 	// +kubebuilder:validation:Minimum=0
 	MinReplicas int32 `json:"minReplicas"`
+
+	// MaxReplicas constrains the maximal number of replicas
 	// +kubebuilder:validation:Minimum=1
-	MaxReplicas    int32                       `json:"maxReplicas"`
+	MaxReplicas int32 `json:"maxReplicas"`
```
vikaschoudhary16: can we please add a little more details like the suggesttion in above comment.

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -44,11 +57,22 @@ type ClusterAutoscalerList struct {
...
+	// Minimum and maximum number of cores in cluster, in the format :.
+	// Cluster autoscaler will not scale the cluster beyond these numbers.
+	Cores *ResourceRange `json:"cores,omitempty"`
+
+	// Minimum and maximum number of gigabytes of memory in cluster, in the format :.
+	// Cluster autoscaler will not scale the cluster beyond these numbers.
+	Memory *ResourceRange `json:"memory,omitempty"`
+
+	// Minimum and maximum number of different GPUs in cluster, in the format <gpu_type>::.
+	// Cluster autoscaler will not scale the cluster beyond these numbers. Can be passed multiple times.
```
vikaschoudhary16: is this correct `in the format <gpu_type>::.` ?

install/01_clusterautoscaler.crd.yaml
```diff
@@ -2,37 +2,52 @@ apiVersion: apiextensions.k8s.io/v1beta1
...
+                trigger Cluster Autoscaler actions, but only run when there are spare
+                resources available, More info: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-does-cluster-autoscaler-work-with-pod-priority-and-preemption'
               format: int32
               type: integer
             resourceLimits:
+              description: Constraints of autoscaling resources
               properties:
                 cores:
+                  description: Minimum and maximum number of cores in cluster, in
+                    the format :. Cluster autoscaler will not scale the cluster beyond
```
vikaschoudhary16: in the format?

Gopkg.toml
bison: I think we should keep it in `vendor` and add a step to the build that re-generates the CRDs and fails if a diff is found.

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
ingvagabund: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca

install/01_clusterautoscaler.crd.yaml
ingvagabund: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca

pkg/apis/autoscaling/v1beta1/machineautoscaler_types.go
ingvagabund: Maybe `MinReplicas constrains the minimal number of replicas of a scalable resource`?

install/01_clusterautoscaler.crd.yaml
ingvagabund: changed to `<min>:<max>`

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
ingvagabund: changed to `<gpu_type>:<min>:<max>`

pkg/apis/autoscaling/v1beta1/machineautoscaler_types.go
ingvagabund: done

pkg/apis/autoscaling/v1beta1/machineautoscaler_types.go
ingvagabund: done

hack/gen-crd.sh
ingvagabund: uncommented

Gopkg.toml
ingvagabund: I am gonna ignore this discussion then

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -68,13 +92,22 @@ type ResourceRange struct {
 }
 
 type ScaleDownConfig struct {
+	// Should CA scale down the cluster
 	Enabled bool `json:"enabled"`
+
+	// How long after scale up that scale down evaluation resumes
```
frobware: Do we want to be more restrictive than a-z? It is supposed to be a duration.

hack/gen-crd.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -eu
+
+echo "Building controller-gen tool..."
+go build -o bin/controller-gen github.com/openshift/cluster-autoscaler-operator/vendor/sigs.k8s.io/controller-tools/cmd/controller-gen
+
+bin/controller-gen crd --domain openshift.io
+
+echo "Coping generated CRDs"
```
frobware: nit: Copying.

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
ingvagabund: It's copy pasted from the upstream documentation.

hack/gen-crd.sh
ingvagabund: done

PR #100, opened 2019-05-14
Use tls-reload branch of controller-runtime

bison:
Updating controller-runtime dependency to use the tls-reload branch temporarily.


hack/e2e.sh
```diff
@@ -0,0 +1,18 @@
...
+set -euo pipefail
+
+GOPATH="$(mktemp -d)"
+export GOPATH
+
+ACTUATOR_PKG="github.com/openshift/cluster-api-actuator-pkg"
+
+go get -u -d "${ACTUATOR_PKG}/..."
+
+go test -timeout 60m \
```
enxebre: In a follow up we'll want to consolidate on `actuator-pkg make test-e2e` instead

PR #101, opened 2019-05-15
Add support for admission webhooks

bison:
This adds support for running validating / mutating admission
webhooks.  The configuration for the webhooks is expected to be
applied by the cluster-version-operator, and TLS assets are generated
by the service-ca-operator.

The service-ca-operator does not yet support injecting the CA bundle
into webhook configurations, so for now the operator reads the CA at
start-up and patches the webhook configuration resource with the CA.

This updates controller-runtime to latest master which includes support for automatically updating the TLS configuration of the webhook server. It also updates to Kubernetes 1.14.0 and removes the now unused copy of cluster-api-actuator-pkg from vendor.


ingvagabund:
Master branch is open for new changes, the PR does not break anything, the code is self-explanatory as long as my knowledge goes. Let's merge it and see how far we can get.

/approve

PR #102, opened 2019-06-21
Enable autoscaler --balance-similar-node-groups option

frobware:
Add support for enabling the --balance-similar-node-groups
cluster-autoscaler option. It defaults to disabled but can be enabled
as follows:

```yaml
apiVersion: "autoscaling.openshift.io/v1"
kind: "ClusterAutoscaler"
metadata:
  name: "default"
spec:
  balanceSimilarNodeGroups: true
```

### DIFF
--- a/pkg/apis/autoscaling/v1/clusterautoscaler_types.go
+++ b/pkg/apis/autoscaling/v1/clusterautoscaler_types.go
@@ -23,6 +23,13 @@ type ClusterAutoscalerSpec struct {
 	// Cluster Autoscaler actions, but only run when there are spare resources available,
 	// More info: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-does-cluster-autoscaler-work-with-pod-priority-and-preemption
 	PodPriorityThreshold *int32 `json:"podPriorityThreshold,omitempty"`
+
+	// BalanceSimilarNodeGroups enables/disables the
+	// `--balance-similar-node-groups` cluster-autocaler feature.
+	// This feature will automatically identify node groups with
+	// the same instance type and the same set of labels and try
+	// to keep the respective sizes of those node groups balanced.
+	BalanceSimilarNodeGroups *bool `json:"balanceSimilarNodeGroups,omitempty"`
 }
 
 // ClusterAutoscalerStatus defines the observed state of ClusterAutoscaler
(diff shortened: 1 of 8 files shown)


frobware:
/hold

Not sure what the procedure is for adding a new field and upgrades.

pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -81,6 +82,10 @@ func AutoscalerArgs(ca *v1.ClusterAutoscaler, cfg *Config) []string {
 		args = append(args, ScaleDownArgs(s.ScaleDown)...)
 	}
 
+	if ca.Spec.BalanceSimilarNodeGroups {
+		args = append(args, BalanceSimilarNodeGroupsArgs()...)
```
bison: I don't think `BalanceSimilarNodeGroupsArgs` needs to be a function. Only the bits that take a full struct or sub-section of the resource as arugments are functions here. I think in this case you can just do:
```golang
args = append(args, BalanceSimilarNodeGroupsArg.String())
```

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -23,6 +23,13 @@ type ClusterAutoscalerSpec struct {
 	// Cluster Autoscaler actions, but only run when there are spare resources available,
 	// More info: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-does-cluster-autoscaler-work-with-pod-priority-and-preemption
 	PodPriorityThreshold *int32 `json:"podPriorityThreshold,omitempty"`
+
+	// BalanceSimilarNodeGroups enables/disables the
+	// `--balance-similar-node-groups` cluster-autocaler feature.
+	// This feature will automatically identify node groups with
+	// the same instance type and the same set of labels and try
+	// to keep the respective sizes of those node groups balanced.
+	BalanceSimilarNodeGroups bool `json:"balanceSimilarNodeGroups"`
```
bison: To make this truly optional, does this need to be a pointer to a bool with `omitempty`?

frobware: Switched to *ptr.

pkg/controller/clusterautoscaler/clusterautoscaler.go
frobware: Done.

frobware:
@bison I think we both concluded (via slack/demo) that rollback is fine as things stand today.

/hold cancel

PR #104, opened 2019-06-26
Monitoring integration

enxebre:
This enables monitoring integration for cao.
- Controller runtime metrics are exposed for machine_autoscaler_controller and cluster_autoscaler_controller
- This enables any farther metric to be scraped by prometheus for either these controllers or any managed component i.e machine-provider/cluster-autoscaler

### DIFF
--- a/install/03_rbac.yaml
+++ b/install/03_rbac.yaml
@@ -221,3 +221,37 @@ subjects:
   - kind: ServiceAccount
     name: cluster-autoscaler
     namespace: openshift-machine-api
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: prometheus-k8s-cluster-autoscaler-operator
+  namespace: openshift-machine-api
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s-cluster-autoscaler-operator
+subjects:
+  - kind: ServiceAccount
+    name: prometheus-k8s
+    namespace: openshift-monitoring
+
+---
+# Roles needed by prometheus to scrape Cluster Autoscaler Operator metrics endpoint
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: prometheus-k8s-cluster-autoscaler-operator
+  namespace: openshift-machine-api
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - services
+      - endpoints
+      - pods
+    verbs:
+      - get
+      - list
+      - watch
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
@@ -28,7 +28,7 @@ import (
 )
 
 const (
-	controllerName      = "cluster-autoscaler-controller"
+	controllerName      = "cluster_autoscaler_controller"
 	caServiceAccount    = "cluster-autoscaler"
 	caPriorityClassName = "system-cluster-critical"
 )
--- a/pkg/operator/config.go
+++ b/pkg/operator/config.go
@@ -1,10 +1,9 @@
 package operator
 
 import (
+	"fmt"
 	"os"
 	"strconv"
-
-	"k8s.io/klog"
 )
 
 const (
@@ -58,6 +57,9 @@ const (
 	// DefaultWebhooksCertDir is the default directory for admission webhook
 	// server TLS assets.
 	DefaultWebhooksCertDir = "/etc/cluster-autoscaler-operator/tls"
+
+	// DefaultMetricsPort is the default port to expose metrics.
+	DefaultMetricsPort = 8080
 )
 
 // Config represents the runtime configuration for the operator.
@@ -122,6 +124,9 @@ type Config struct {
 	// WebhookCertDir is the directory containing TLS assets for the admission
 	// webhook server.
 	WebhooksCertDir string
+
+	// metricsPort is the port the metrics are exposed.
+	MetricsPort int
 }
 
 // NewConfig returns a new Config object with defaults set.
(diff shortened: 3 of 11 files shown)


pkg/operator/config.go
```diff
@@ -228,5 +235,15 @@ func ConfigFromEnvironment() *Config {
...
 	}
 
+	if metricsPort, ok := os.LookupEnv("METRICS_PORT"); ok {
+		v, err := strconv.ParseInt(metricsPort, 10, 32)
+		if err != nil {
+			v = DefaultMetricsPort
+			klog.Errorf("Error parsing METRICS_PORT environment variable: %v", err)
+		}
+
+		config.metricsPort = int(v)
```
frobware: If the `ParseInt` fails then metricsPort is likely to (the zero value) 0. This will then listen on port 0 -- the ephemeral port number - and will appear to work as the OS will assign an unused port number. I don't think we want this. We should return an error if the conversion fails.

pkg/operator/config.go
```diff
@@ -228,5 +235,15 @@ func ConfigFromEnvironment() *Config {
 		config.WebhooksCertDir = webhooksCertDir
 	}
 
+	if metricsPort, ok := os.LookupEnv("METRICS_PORT"); ok {
+		v, err := strconv.ParseInt(metricsPort, 10, 32)
+		if err != nil {
+			v = DefaultMetricsPort
+			klog.Errorf("Error parsing METRICS_PORT environment variable: %v", err)
```
frobware: Log the value that was parsed, but failed:

`klog.Errorf("Error parsing %q METRICS_PORT environment variable: %v", os.LookupEnv("METRICS_PORT"), err)`

pkg/operator/config.go
```diff
@@ -215,18 +223,26 @@ func ConfigFromEnvironment() *Config {
...
 
 	if webhooksCertDir, ok := os.LookupEnv("WEBHOOKS_CERT_DIR"); ok {
 		config.WebhooksCertDir = webhooksCertDir
 	}
 
-	return config
+	if metricsPort, ok := os.LookupEnv("METRICS_PORT"); ok {
+		v, err := strconv.Atoi(metricsPort)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing METRICS_PORT (%q) environment variable: %v", metricsPort, err)
```
frobware: Let's raise the bar and add unit tests for the new behaviour, testing for both positive/negative results when setting `METRICS_PORT` and `WEBHOOKS_CERT_DIR`.

install/0000_90_cluster-autoscaler-operator_01_prometheus_rbac.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
bison: Why are this file and the ServiceMonitor file named differently from the others? There is an existing file with RBAC rules.

pkg/operator/config_test.go
```diff
@@ -12,3 +16,89 @@ func TestNewConfig(t *testing.T) {
...
+	for _, tc := range testCase {
+		for key, val := range tc.envVars {
+			os.Setenv(key, val)
+		}
+		got, err := ConfigFromEnvironment()
+		if (err != nil) != tc.expectedError {
+			t.Errorf("expected %v, got: %v", tc.expectedError, err)
+		}
+		if !reflect.DeepEqual(got, tc.expectedConfig) {
+			t.Errorf("Got: %v, expected: %v", got, tc.expectedConfig)
```
frobware: Keep the args the same as the previous `t.Errorf()` for consistency (i.e., expected, got).

install/0000_90_cluster-autoscaler-operator_02_servicemonitor.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    k8s-app: cluster-autoscaler-operator
+  name: cluster-autoscaler-operator
+  namespace: openshift-monitoring
```
bison: This seems wrong. Shouldn't this go in our namespace?

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 const (
-	controllerName      = "cluster-autoscaler-controller"
+	controllerName      = "cluster_autoscaler_controller"
```
bison: I know why you're doing this, but I really wish it was fixed in controller-runtime...

pkg/operator/config.go
```diff
@@ -122,6 +124,9 @@ type Config struct {
 	// WebhookCertDir is the directory containing TLS assets for the admission
 	// webhook server.
 	WebhooksCertDir string
+
+	// metricsPort is the port the metrics are exposed.
+	metricsPort int
```
bison: Why would the metrics port be unexported unlike all the other fields?

enxebre: any reason for them to be exported? should we instead unexport in a different PR?

bison: They just seem like public fields to me. This is not something that's going to cross system boundaries. We don't have to be super protective of it, but if I want to pass it to another package within the project at some point, I should be able to read the fields.

enxebre: ok sg

install/03_rbac.yaml
```diff
@@ -221,3 +221,37 @@ subjects:
   - kind: ServiceAccount
     name: cluster-autoscaler
     namespace: openshift-machine-api
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: prometheus-k8s
```
bison: A bit concerned about naming of the Role and RoleBinding. We share a namespace with all the other cluster-api related components. Will they each create this when they have monitoring enabled? If so, they should probably have the operator name included here to create separate resources, even if the permissions are the same.

enxebre: great point

PR #105, opened 2019-06-26
Move management of webhook configurations to operator

bison:
This moves the management of the webhook configurations under the
control of the operator rather than the CVO.  Unfortunately the CA
bundle fields in the webhook configurations are not known until
runtime.  If the CVO manages the configuration resources, the CA field
cannot be updated as the CVO would overwrite any changes.


install/03_rbac.yaml
```diff
@@ -10,10 +10,7 @@ rules:
   - validatingwebhookconfigurations
   - mutatingwebhookconfigurations
   verbs:
-  - get
-  - patch
-  resourceNames:
-  - autoscaling.openshift.io
+  - "*"
```
michaelgugino: Do we need mutating webhook permissions?

pkg/operator/webhookconfig.go
```diff
@@ -1,84 +1,121 @@
...
 }
 
+// ValidatingWebhooks returns the validating webhook configurations.
+func (w *WebhookConfigUpdater) ValidatingWebhooks() ([]admissionregistrationv1beta1.Webhook, error) {
+	caBundle, err := w.GetEncodedCA()
+	if err != nil {
+		return nil, err
+	}
+
+	failurePolicy := admissionregistrationv1beta1.Ignore
```
michaelgugino: Why should this be ignore?

install/03_rbac.yaml
bison: Not yet. We can remove that for now. Done.

pkg/operator/webhookconfig.go
bison: Because the validations are really (or will be when we're not doing allow-all) a nice-to-have. We have OpenAPI validation on the fields where we're able to do that, and other errors will be reported in events or in status conditions after the resource is created. So, we shouldn't completely block the functioning of the operator if the connection to the webhook server is not working -- e.g. if there's some transient networking issue.

enxebre:
more context https://coreos.slack.com/archives/C68TNFWA2/p1561560240498300

enxebre:
This will only create the webhook once and won't take care of reconciling/recreation. If we find no way to drive this via cvo we'll need to make it more robust via local controller soon.
/approve

PR #107, opened 2019-06-28
Add validator for ClusterAutoscaler resources

bison:
This adds a validator type for ClusterAutoscaler resources.  It handles validation both inside the reconciliation loop, where it returns early and records an event on failure, and for the validating webhook.

Here's an example of the reporting users now get earlier in the flow of deploying the cluster-autoscaler -- here the maximum number of cores was lower than the minimum and one of the durations in the scale down configuration was invalid:

```console
$ kubectl apply -f clusterautoscaler.yaml 
Error from server ([ResourceLimits.Cores: Max must be greater than or equal to Min, ScaleDown.DelayAfterAdd: time: invalid duration bad-duration]): error when creating "clusterautoscaler.yaml": admission webhook "clusterautoscalers.autoscaling.openshift.io" denied the request: [ResourceLimits.Cores: Max must be greater than or equal to Min, ScaleDown.DelayAfterAdd: time: invalid duration bad-duration]
```

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
@@ -34,12 +34,13 @@ const (
 )
 
 // NewReconciler returns a new Reconciler.
-func NewReconciler(mgr manager.Manager, cfg *Config) *Reconciler {
+func NewReconciler(mgr manager.Manager, config Config) *Reconciler {
 	return &Reconciler{
-		client:   mgr.GetClient(),
-		scheme:   mgr.GetScheme(),
-		recorder: mgr.GetEventRecorderFor(controllerName),
-		config:   cfg,
+		client:    mgr.GetClient(),
+		scheme:    mgr.GetScheme(),
+		recorder:  mgr.GetEventRecorderFor(controllerName),
+		validator: NewValidator(config.Name),
+		config:    config,
 	}
 }
 
@@ -69,10 +70,11 @@ var _ reconcile.Reconciler = &Reconciler{}
 type Reconciler struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
-	client   client.Client
-	recorder record.EventRecorder
-	scheme   *runtime.Scheme
-	config   *Config
+	client    client.Client
+	recorder  record.EventRecorder
+	config    Config
+	scheme    *runtime.Scheme
+	validator *Validator
 }
 
 // AddToManager adds a new Controller to mgr with r as the reconcile.Reconciler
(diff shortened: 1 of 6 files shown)


ingvagabund:
@bison worth writing a simple e2e test that validates an invalid CA is refused and the resource is not created at all. Given `go get` does not allow to specify a git branch, we need to merge this PR first. Then, write new e2e test for the new functionality.

/approve

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -149,6 +151,15 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	// generated for these cluster scoped objects out of the default namespace.
 	caRef := r.objectReference(ca)
 
+	// Validate the ClusterAutoscaler early and requeue if any errors are found.
```
frobware: Can `r.validate` ever be `nil` at this point?

pkg/controller/clusterautoscaler/validator.go
```diff
@@ -15,6 +19,125 @@ import (
...
+
+// validateScaleDownConfig validates ScaleDownConfig objects.
+func (v *Validator) validateScaleDownConfig(sd *autoscalingv1.ScaleDownConfig) utilerrors.Aggregate {
+	var errs []error
+
+	durations := map[string]*string{
+		"DelayAfterAdd":     sd.DelayAfterAdd,
+		"DelayAfterDelete":  sd.DelayAfterDelete,
+		"DelayAfterFailure": sd.DelayAfterFailure,
+		"UnneededTeim":      sd.UnneededTime,
```
frobware: Typo: `UnneededTime`.

pkg/controller/clusterautoscaler/validator.go
```diff
@@ -27,8 +150,12 @@ func (v *Validator) Handle(ctx context.Context, req admission.Request) admission
 
 	klog.Infof("Validation webhook called for ClustAutoscaler: %s", ca.GetName())
 
+	if ok, err := v.Validate(ca); !ok {
+		return admission.Denied(err.Error())
+	}
+
 	// TODO: Implement validations.
```
enxebre: Remove TODO?

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
bison: Only if you didn't use `NewReconciler()` and purposefully constructed your own reconciler without a validator. The same goes for all the fields though, e.g. the manager, client, etc.

pkg/controller/clusterautoscaler/validator.go
bison: Oops. Fixed.

pkg/controller/clusterautoscaler/validator.go
bison: Oops. Fixed.

PR #108, opened 2019-07-02
Add validator for MachineAutoscaler resources

bison:
This adds a validator type for MachineAutoscaler resources. It handles
validation both inside the reconciliation loop, where it returns early
and records an event on failure, and for the validating webhook.

### DIFF
--- a/pkg/controller/machineautoscaler/validator.go
+++ b/pkg/controller/machineautoscaler/validator.go
@@ -1,0 +1,77 @@
+package machineautoscaler
+
+import (
+	"context"
+	"errors"
+	"net/http"
+
+	autoscalingv1beta1 "github.com/openshift/cluster-autoscaler-operator/pkg/apis/autoscaling/v1beta1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// Validator validates MachineAutoscaler resources.
+type Validator struct {
+	client  client.Client
+	decoder *admission.Decoder
+}
+
+// NewValidator returns a new Validator.
+func NewValidator() *Validator {
+	return &Validator{}
+}
+
+// Validate validates the given MachineAutoscaler resource.
+func (v *Validator) Validate(ma *autoscalingv1beta1.MachineAutoscaler) (bool, utilerrors.Aggregate) {
+	var errs []error
+
+	if ma == nil {
+		err := errors.New("MachineAutoscaler is nil")
+		return false, utilerrors.NewAggregate([]error{err})
+	}
+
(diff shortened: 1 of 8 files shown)


pkg/controller/machineautoscaler/validator.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// NewValidator returns a new Validator.
+func NewValidator() *Validator {
+	return &Validator{}
+}
+
+// Validate validates the given MachineAutoscaler resource.
+func (v *Validator) Validate(ma *autoscalingv1beta1.MachineAutoscaler) (bool, utilerrors.Aggregate) {
+	errs := []error{}
```
frobware: Replace with nil slice declaration? - `var errs []error`

bison: I'm happy to, but any reason? Seems half of people prefer one way and the other half the other. Doesn't make any difference to me.

frobware: I just try to be consistent with the implied semantics, ala:

```go
	var nilSlice []int
	emptySlice1 := make([]int, 0)
	emptySlice2 := []int{}

	fmt.Println(nilSlice == nil)    // true
	fmt.Println(emptySlice1 == nil) // false
	fmt.Println(emptySlice2 == nil) // false
```

frobware: Not a blocker for me  - just something I try to be self-consistent with and that's why I noticed it.

bison: Sure. Makes sense.

pkg/controller/machineautoscaler/validator.go
```diff
@@ -0,0 +1,77 @@
...
+	ma := &autoscalingv1beta1.MachineAutoscaler{}
+
+	if err := v.decoder.Decode(req, ma); err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	klog.Infof("Validation webhook called for MachineAutoscaler: %s", ma.GetName())
+
+	if ok, err := v.Validate(ma); !ok {
+		return admission.Denied(err.Error())
```
frobware: Is `err` always guaranteed to be non-`nil` here? Otherwise, boom!

bison: Yes. We only return `ok == true` if there is at least one error, which means this will never be nil if `!ok`. I suppose it wouldn't hurt to add that to the condition here though.

pkg/controller/machineautoscaler/validator.go
bison: Fixed.

pkg/controller/machineautoscaler/validator.go
bison: Added a check in the unit tests that fails if `!ok && err == nil`.

enxebre:
is there a counter part PR for the e2e suite?
/approve

bison:
@enxebre, yeah it's mentioned above in the timeline: https://github.com/openshift/cluster-api-actuator-pkg/pull/94

PR #109, opened 2019-07-03
Fix insecure test certificate host names

bison:
Fixes https://github.com/openshift/cluster-api-actuator-pkg/pull/94. The API server was using a DNS name for the service that was not present in the certificate.


ingvagabund:
silently approving
/approve

PR #110, opened 2019-07-04
Autoscaler metrics and alerts

enxebre:
Expose the metrics port on the autoscaler pod
Integrate autoscaler with prometheus via serviceMonitor
Create generic alerts for autoscaler


bison:
This deploys the Service, ServiceMonitor, and alerting rules unconditionally, but the cluster-autoscaler itself is not deployed by default. Is that a problem? At minimum, it means there will be a target in Prometheus that reports as down on every cluster until you actually deploy the autoscaler.

The alternative would be to have the operator deploy these things, or at least the Service + ServiceMontior, when a ClusterAutoscaler resource is created.

frobware:
> The alternative would be to have the operator deploy these things, or at least the Service + ServiceMontior, when a ClusterAutoscaler resource is created.

This seems cleaner to me.

vikaschoudhary16:
yeah, objects for cluster-autoscaler like service etc, if deployed by operator instead through bindata, that would be more cleaner

enxebre:
I'd prefer to keep these resources under cvo control so we avoid specific operator lifecycle management opinions, duplicated logic and granting additional perms. @smarterclayton @derekwaynecarr @brancz do we have a convention here, is it a tragedy to have the serviceMonitor deployed even when there's no instance of the autoscaler running?

install/09_alertrules.yaml
```diff
@@ -0,0 +1,38 @@
...
+            severity: critical
+          annotations:
+            message: "Cluster Autoscaler has {{ $value }} unready nodes"
+
+        - alert: ClusterAutoscalerNotSafeToScale
+          expr: |
+            cluster_autoscaler_cluster_safe_to_autoscale{service="cluster-autoscaler"} != 1
+          for: 15m
+          labels:
+            severity: critical
```
brancz: All of these alerts are cause based, not symptom based. Paging should happen on symptom based factors such as "Users can't schedule pods", these all seem like "warning" type alerts at most, maybe even "info" level.

install/07_servicemonitor.yaml
```diff
@@ -17,3 +17,24 @@ spec:
...
+  namespace: openshift-machine-api
+spec:
+  endpoints:
+    - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      interval: 30s
+      port: metrics
+      scheme: http
+  namespaceSelector:
+    matchNames:
+      - openshift-machine-api
```
brancz: This defaults to the namespace the ServiceMonitor is created in, so this is unnecessary.

brancz:
> is it a tragedy to have the serviceMonitor deployed even when there's no instance of the autoscaler running

It's not a tragedy as long as something _will_ appear. To ensure this, you can add absent style alert like these: https://github.com/openshift/cluster-monitoring-operator/blob/30ed3a0e04b851be8b349d9e64fcdeb5d950dff5/assets/prometheus-k8s/rules.yaml#L436-L443

bison:
> It's not a tragedy as long as something will appear.

Yeah, that's the problem. Nothing is guaranteed to appear. It only will if the user manually creates a ClusterAutoscaler resource. So it's probably better if this stuff is deployed by the operator as part of that process.

@brancz, is there any problem with having the CVO deploy the alerting rules, but only deploy the ServiceMonitor when the user creates a ClusterAutoscaler? Will that result in alerts firing?

brancz:
> @brancz, is there any problem with having the CVO deploy the alerting rules, but only deploy the ServiceMonitor when the user creates a ClusterAutoscaler? Will that result in alerts firing?

Yes it will, but I misunderstood the situation, I don't think that's what you want.

> Yeah, that's the problem. Nothing is guaranteed to appear. It only will if the user manually creates a ClusterAutoscaler resource. So it's probably better if this stuff is deployed by the operator as part of that process.

Agreed, I think this is the right choice here.

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -35,6 +36,7 @@ const (
 
 // NewReconciler returns a new Reconciler.
 func NewReconciler(mgr manager.Manager, config Config) *Reconciler {
+	monitoringv1.AddToScheme(mgr.GetScheme())
```
bison: I would probably just do this where we instantiate the operator and manager: [here](https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/operator/operator.go#L65)

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
```diff
@@ -183,6 +184,7 @@ func TestCanGetca(t *testing.T) {
 // newFakeReconciler returns a new reconcile.Reconciler with a fake client
 func newFakeReconciler(initObjects ...runtime.Object) *Reconciler {
 	fakeClient := fakeclient.NewFakeClient(initObjects...)
+	monitoringv1.AddToScheme(scheme.Scheme)
```
bison: Maybe just do this in `init()` like the other APIs.

pkg/controller/clusterautoscaler/monitoring.go
```diff
@@ -0,0 +1,258 @@
...
+// for the given ClusterAutoscaler custom resource instance.
+func (r *Reconciler) ensureAutoscalerService(ca *autoscalingv1.ClusterAutoscaler) error {
+	desired := r.AutoscalerService(ca)
+	// Set ClusterAutoscaler instance as the owner and controller.
+	if err := controllerutil.SetControllerReference(ca, desired, r.scheme); err != nil {
+		return err
+	}
+
+	current := &corev1.Service{}
+	if err := r.client.Get(context.TODO(), r.AutoscalerName(ca), current); err != nil {
```
bison: I think GitHub ate my previous comment because of a force push, so sorry if this shows up twice.

Would it be easier in these functions to just use [CreateOrUpdate](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrUpdate)?

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
frobware: Can we do this elsewhere?

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -115,6 +116,22 @@ func (r *Reconciler) AddToManager(mgr manager.Manager) error {
 		OwnerType:    &autoscalingv1.ClusterAutoscaler{},
 	})
 
+	// Watch for changes to monitoring resources owned by a ClusterAutoscaler
+	err = c.Watch(&source.Kind{Type: &corev1.Service{}}, &handler.EnqueueRequestForOwner{
```
frobware: We never check `err` for the next 3 watches.

pkg/controller/clusterautoscaler/monitoring.go
```diff
@@ -0,0 +1,210 @@
...
+		return "", err
+	}
+
+	klog.V(4).Infof("Ensuring object %q of type %T, operation: %v", desired.GetName(), desired, op)
+	return op, nil
+}
+
+// ensureAutoscalerServiceMonitor will create a service
+// for the given ClusterAutoscaler custom resource instance.
+func (r *Reconciler) ensureAutoscalerService(ca *autoscalingv1.ClusterAutoscaler) (controllerutil.OperationResult, error) {
```
frobware: We use "ensure" but I think it would be clearer if it hinted at what it attempts to "ensure" - `ensureAutoscalerServiceMonitor{Exists,Created,SomethingElse}`? Ditto for the other following ensure-based functions. Edit: having looked at the related functions do we want to s/ensure/create, or s/ensure/createOrUpdate?

pkg/controller/clusterautoscaler/monitoring.go
```diff
@@ -0,0 +1,210 @@
...
+	if _, err := r.ensureAutoscalerService(ca); err != nil {
+		return fmt.Errorf("error ensuring cluster autoscaler service: %v", err)
+	}
+
+	if _, err := r.ensureAutoscalerServiceMonitor(ca); err != nil {
+		return fmt.Errorf("error ensuring cluster autoscaler serviceMonitor: %v", err)
+	}
+
+	if _, err := r.ensureAutoscalerPrometheusRule(ca); err != nil {
+		return fmt.Errorf("error ensuring cluster autoscaler serviceMonitor: %v", err)
```
frobware: typo: "cluster autoscaler Prometheus rules"?

PR #111, opened 2019-07-05
README: Add info on webhook infrastructure

bison:
Adds info to the README on validating webhooks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,51 @@ deployment for the operator.
 ### End-to-End Tests
 
 You can run the e2e test suite with `make test-e2e`.  These tests
-assume the presence of a cluster not already running the operator, and
+assume the presence of a cluster already running the operator, and
 that the `KUBECONFIG` environment variable points to a configuration
 granting admin rights on said cluster.
+
+
+## Validating Webhooks
+
+By default the operator starts an HTTP server for webhooks and
+registers a `ValidatingWebhookConfiguration` with the API server for
+both the `ClusterAutoscaler` and `MachineAutoscaler` types.  This can
+be disabled via the `WEBHOOKS_ENABLED` environment variable.  At the
+moment, reconciliation of the webhook configuration is only performed
+once at startup after leader-election has succeeded.
+
+If the webhook server is enabled, you must provide a TLS certificate
+and key as well as a CA certificate to the operator.  The location of
+these is controlled by the `WEBHOOKS_CERT_DIR` environment variable,
+which defaults to: `/etc/cluster-autoscaler-operator/tls`
+
+The files must be in the following locations:
+
+  - `${WEBHOOKS_CERT_DIR}/tls.crt`
+  - `${WEBHOOKS_CERT_DIR}/tls.key`
+  - `${WEBHOOKS_CERT_DIR}/service-ca/ca-cert.pem`
+
+The default cluster-autoscaler-operator deployment on OpenShift will
+generate the TLS assets automatically with the help of the OpenShift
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,6 +79,48 @@ deployment for the operator.
...
+the operator pod.  Additionally, the service-ca-operator injects its
+CA certificate into a `ConfigMap`, which is also mounted.  The
+operator then uses the TLS certificate and key to secure the webhook
+HTTP server, and injects the CA certificate into the webhook
+configuration registered with the API server.
+
+Rotation of the TLS certificate and key is handled transparently.  The
+[controller-runtime][controller-runtime] library the operator is based
+on watches the files mounted in the pod for changes and updates HTTP
+server's TLS configuration.  Rotation of the CA certificate is
```
frobware: What's the implication (if any) of not handling rotation?

vikaschoudhary16: what is the frequency of tls cert and key automatic rotation at webhook server? should this be disabled, otherwise will not it create a problem because ca crt is not updated automatically?

frobware: > Well, it depends, but if the key changes then all the certificates will be updated as well. The webhook server will get its new certificate and key, but the API server will not be able to validate them until the CA is updated, which only happens at startup. This isn't really a problem in practice at the moment though, because: A) The CA is not short-lived. B) The webhook config is set to ignore failure as the validation is just a convenience. C) The CA will be updated after a restart any continue working anyway.

Can we add that to the doc, perhaps under a section `Cert Rotation`?

bison: Sure, I'll try to summarize it. This is really a short summary of the behavior though. I don't want it to turn unto an explanation of the mechanics of rotating TLS certificates.

frobware: Agreed. But given that we have that explanation it seems useful to capture it and not lose it to this GH PR conversation.

README.md
```diff
@@ -79,6 +79,48 @@ deployment for the operator.
...
 granting admin rights on said cluster.
+
+
+## Validating Webhooks
+
+By default the operator starts an HTTP server for webhooks and
+registers a `ValidatingWebhookConfiguration` with the API server for
+both the `ClusterAutoscaler` and `MachineAutoscaler` types.  This can
+be disabled via the `WEBHOOKS_ENABLED` environment variable.  At the
+moment, reconciliation of the webhook configuration is only preformed
```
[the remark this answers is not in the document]
bison: Thanks. Fixed.

README.md
```diff
@@ -79,6 +79,48 @@ deployment for the operator.
...
+these is controlled by the `WEBHOOKS_CERT_DIR` environment variable,
+which defaults to: `/etc/cluster-autoscaler-operator/tls`
+
+The files must be in the following locations:
+
+  - `${WEBHOOKS_CERT_DIR}/tls.crt`
+  - `${WEBHOOKS_CERT_DIR}/tls.key`
+  - `${WEBHOOKS_CERT_DIR}/service-ca/ca-cert.pem`
+
+The default deployment on OpenShift will generate the TLS assets for
```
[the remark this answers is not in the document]
bison: Fixed.

README.md
bison: Updated.

> what is the frequency of tls cert and key automatic rotation at webhook server? should this be disabled, otherwise will not it create a problem because ca crt is not updated automatically?

I'm not sure on the frequency. On OpenShift, that's up to the service-ca-operator. We should definitely not disable rotation of the TLS certificate. It can only cause a problem if the CA also changes, and then the only problem is that validations happen a bit later AND it will start working again on the next restart.

PR #112, opened 2019-07-27
Enable ignore-daemonsets-utilization and skp-nodes-with-local-storage

vikaschoudhary16:
These options are being used in 3.11. This PR enables these options and makes it possible to be able to set these at the cluster autoscaler operator.
https://jira.coreos.com/browse/CLOUD-603

### DIFF
--- a/pkg/apis/autoscaling/v1/clusterautoscaler_types.go
+++ b/pkg/apis/autoscaling/v1/clusterautoscaler_types.go
@@ -31,9 +31,6 @@ type ClusterAutoscalerSpec struct {
 	// to keep the respective sizes of those node groups balanced.
 	BalanceSimilarNodeGroups *bool `json:"balanceSimilarNodeGroups,omitempty"`
 
-	// Enables/Disables `--write-status-configmap` CA feature flag. Should CA write status information to a configmap. true by default
-	WriteStatusConfigMap *bool `json:"writeStatusConfigMap,omitempty"`
-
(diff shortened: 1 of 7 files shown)


pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -30,6 +30,15 @@ type ClusterAutoscalerSpec struct {
 	// the same instance type and the same set of labels and try
 	// to keep the respective sizes of those node groups balanced.
 	BalanceSimilarNodeGroups *bool `json:"balanceSimilarNodeGroups,omitempty"`
+
+	// Enables/Disables `--write-status-configmap` CA feature flag. Should CA write status information to a configmap. true by default
+	WriteStatusConfigMap *bool `json:"writeStatusConfigMap,omitempty"`
```
derekwaynecarr: if its true by default, what is the use case for disabling it?

vikaschoudhary16: not sure. Digging through git history found this commit which adds this option:
https://github.com/kubernetes/autoscaler/commit/84f19c1e1e678b031a256322386e5687d25091ea
Could not find any discussion related to why make it optional and when it could be set to disable.

enxebre: based on the sig-autoscaler conversation I think we should drop it. Only case I can think of is the role for the pod running the ca missing configMap perms

frobware:
@vikaschoudhary16  Please could you update the README to cover the additional flags.

PR #113, opened 2019-07-31
Bug 1733471: Fix MachineAutoscaler target updates

bison:
This moves the handling of changes to MachineAutoscaler targets to
early in the reconciliation process -- just after handling of
deletion.  Otherwise, it's possible for the previous target to never
have its autoscaling configuration removed if there's some problem
fetching the new target.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1733471

### DIFF
--- a/pkg/controller/machineautoscaler/machineautoscaler_controller_test.go
+++ b/pkg/controller/machineautoscaler/machineautoscaler_controller_test.go
@@ -1,22 +1,55 @@
 package machineautoscaler
 
 import (
+	"context"
 	"reflect"
+	"strconv"
 	"testing"
 
 	"github.com/openshift/cluster-autoscaler-operator/pkg/apis"
+	autoscalingv1beta1 "github.com/openshift/cluster-autoscaler-operator/pkg/apis/autoscaling/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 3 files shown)


pkg/controller/machineautoscaler/machineautoscaler_controller_test.go
```diff
@@ -142,3 +175,146 @@ func TestValidateReference(t *testing.T) {
...
+				if err := r.client.Get(context.TODO(), maName, ma); err != nil {
+					t.Errorf("Error fetching MachineAutoscaler: %v", err)
+				}
+			}
+
+			// Now set the new target and reconcile again.
+			setTarget(ma, tt.newTarget)
+
+			if err := r.client.Update(context.TODO(), ma); err != nil {
+				t.Errorf("Error updating MachineAutoscaler: %v", err)
```
frobware: Can the remainder of the test continue if Update() fails?

bison: I suppose not. We could make this fatal. Though, in practice, with the fake client, this will never fail.

bison: Fixed. Errors on update or fetch are fatal now.

enxebre:
/approve
I might want to also consider stop allowing a target change pre-presistence

bison:
> I might want to also consider stop allowing a target change pre-presistence

Does that mean just making the target immutable? If so, I'd love to do that, but it's not really possible with CRDs.

The best we could do is disallow updates to the target in the validating webhook, but right now we're considering that validation a convenience and configuring the API server to fail open if it can't talk to the webhook server. So we have to handle this still in case the webhook doesn't catch it. We could configure the API server to fail hard if there's a problem with the webhook, but that comes with its own set of problems.

We could also detect a change to the target and just throw errors, but in that case we might still want to remove the config from the old target. In that case we've already done 80% of the work we're doing here.

PR #118, opened 2019-09-05
Switch to go modules

spangenberg:
Migrated using `go mod init github.com/openshift/cluster-autoscaler-operator`.
Adapting hack scripts to build properly.


spangenberg:
CI Go version is too old to support go modules.
Version was bumped in https://github.com/openshift/aos-cd-jobs/pull/1949
Will be retried once CI it's merged.

ingvagabund:
```
hack/ci-integration.sh  -ginkgo.v -ginkgo.noColor=true -ginkgo.focus "Feature:Operators" -ginkgo.failFast
# _/tmp/tmp.rpaj1p7J7m/pkg/e2e
pkg/e2e/e2e_test.go:6:2: cannot find package "github.com/golang/glog" in any of:
	/usr/local/go/src/github.com/golang/glog (from $GOROOT)
	/go/src/github.com/golang/glog (from $GOPATH)
FAIL	_/tmp/tmp.rpaj1p7J7m/pkg/e2e [setup failed]
```

vendor/github.com/blang/semver/range.go
```diff
@@ -327,7 +327,7 @@ func expandWildcardVersion(parts [][]string) ([][]string, error) {
 	for _, p := range parts {
 		var newParts []string
 		for _, ap := range p {
-			if strings.Contains(ap, "x") {
+			if strings.Index(ap, "x") != -1 {
```
michaelgugino: It seems there are two or more libraries being updated here, not just adding/removing files we didn't use.

enxebre: ping @spangenberg

Dockerfile
```diff
@@ -3,7 +3,7 @@ WORKDIR /go/src/github.com/openshift/cluster-autoscaler-operator
 COPY . .
 ENV NO_DOCKER=1
 ENV BUILD_DEST=/go/bin/cluster-autoscaler-operator
-RUN unset VERSION && make build
+RUN unset VERSION && GOPROXY=off make build
```
enxebre: @spangenberg why do we need to set `GOPROXY=off` explicitly?

hack/e2e.sh
```diff
@@ -2,11 +2,9 @@
 
 set -euo pipefail
 
-GOPATH="$(mktemp -d)"
-export GOPATH
+unset GOFLAGS
```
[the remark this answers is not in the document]
enxebre: @spangenberg I'm still confusing why we need `unset GOFLAGS`, can you elaborate?

Dockerfile
enxebre: Also wouldn't `build` need to be aware of `GO111MODULE` and `GOFLAGS`

hack/e2e.sh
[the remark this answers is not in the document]
spangenberg: `hack/e2e.sh` is executing the `Makefile` in [github.com/openshift/cluster-api-actuator-pkg](https://github.com/openshift/cluster-api-actuator-pkg) which does `GOFLAGS += -mod=vendor` just like this repos `Makefile` does.
This would result in `GOFLAGS=-mod=vendor-mod=vendor` which gets passed to each `go build` command which is wrong. So I unset it here because it get's se again by the other `Makefile`.

Dockerfile
spangenberg: The `Makefile` is setting `GO111MODULE=on` and `GOFLAGS=...`.
`GOPROXY=off` is done here to really make sure it's not fetching anything else and only uses vendored packages in order to guarantee reproducible builds that are hermetic.

hack/e2e.sh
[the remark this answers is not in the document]
enxebre: thanks! is there a reason to concat rather than just set?

Makefile
```diff
@@ -24,14 +31,12 @@ else
   IMAGE_BUILD_CMD = docker build
 endif
 
-.PHONY: depend
```
enxebre: wouldn't `DOCKER_CMD` above need to be env `GO111MODULE` and `GOFLAGS` aware?

Dockerfile
enxebre: thanks! so is `GOPROXY` used by go build? my understanding was that it will just look at the vendored folder by setting `GO111MODULE` and -mod=vendor

hack/e2e.sh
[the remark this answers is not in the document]
spangenberg: In case tooling sets other commands with GOFLAGS, that would overwrite it.
Found that patters in other projects too.

Dockerfile
spangenberg: It does, but in case there is something not vendored yet. Because someone forgot to vendor it. Go build would fetch the missing one live from remote.
This way we fail fast.

hack/e2e.sh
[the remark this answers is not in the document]
enxebre: but this is exactly that case and it breaks because it wrongly concat with no spaces, right?

Makefile
enxebre: ping @spangenberg

spangenberg: Thanks, should be fixed now.

PR #120, opened 2019-10-29
Add csinodes to autoscaler ClusterRole

enxebre:
This requirement was introduced by https://github.com/kubernetes/autoscaler/pull/2305 and fixed for e.g aws here https://github.com/kubernetes/autoscaler/pull/2404


enxebre:
We need this for https://github.com/openshift/kubernetes-autoscaler/pull/120 to go green

PR #119, opened 2019-10-28
Secure metrics server

spangenberg:

PR #121, opened 2019-11-12
Drop alert temporary

enxebre:
This is breaking all CI due to alert being trieggered https://github.com/openshift/cluster-autoscaler-operator/pull/119


enxebre:
cc @ingvagabund

PR #122, opened 2019-11-12
Fix metrics

enxebre:


enxebre:
upgrade failing because of ```E1113 13:52:35.285449       1 task.go:77] error running apply for deployment "openshift-machine-api/cluster-autoscaler-operator" (178 of 489): Deployment.apps "cluster-autoscaler-operator" is invalid: spec.template.spec.containers[0].ports[1].hostPort: Duplicate value: "TCP//9192"```

PR #133, opened 2020-02-28
BUG 1806438: Drop run-level

enxebre:
cao should follow normal admission control rules for a cluster.
Needs https://github.com/openshift/machine-api-operator/pull/502

### DIFF
--- a/install/00_namespace.yaml
+++ b/install/00_namespace.yaml
@@ -5,6 +5,5 @@ metadata:
   name: openshift-machine-api
   labels:
     name: openshift-machine-api
-    openshift.io/run-level: "1"
     # allow openshift-monitoring to look for ServiceMonitor objects in this namespace
     openshift.io/cluster-monitoring: "true"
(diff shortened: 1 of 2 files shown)


elmiko:
lgtm

markmc:
@enxebre is this going to break the baremetal pod deployment? It uses hostNetwork, hostPort, and privileged

/cc @sadasu @stbenjam @hardys

enxebre:
@markmc good catch, I believe adding the specifics SCC to the clusterRole used by the `name: machine-api-controllers` should be enough as per https://github.com/openshift/cluster-ingress-operator/commit/1409f54a17d7ee60cd60d4f6e4da8088cf82962c
i.e
```
- apiGroups:
  - security.openshift.io
  resources:
  - securitycontextconstraints
  verbs:
  - use
  resourceNames:
  - hostnetwork
  - privileged
```
cc @Miciah @smarterclayton can you confirm?

/hold

PR #134, opened 2020-03-02
Bug 1809325: ROKS - remove additional autoscaler manifests

csrwng:
Remove prometheus rule and autoscaler CRD that should not be present in a hosted
control plane deployment.

### DIFF
--- a/install/01_clusterautoscaler.crd.yaml
+++ b/install/01_clusterautoscaler.crd.yaml
@@ -5,6 +5,8 @@ metadata:
   labels:
     controller-tools.k8s.io: "1.0"
   name: clusterautoscalers.autoscaling.openshift.io
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
 spec:
   group: autoscaling.openshift.io
   names:
--- a/install/02_machineautoscaler.crd.yaml
+++ b/install/02_machineautoscaler.crd.yaml
@@ -5,6 +5,8 @@ metadata:
   labels:
     controller-tools.k8s.io: "1.0"
   name: machineautoscalers.autoscaling.openshift.io
+  annotations:
+    exclude.release.openshift.io/internal-openshift-hosted: "true"
 spec:
   additionalPrinterColumns:
   - JSONPath: .spec.scaleTargetRef.kind
--- a/hack/gen-crd.sh
+++ b/hack/gen-crd.sh
@@ -2,8 +2,18 @@
 
 set -eu
 
-go run ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen crd --domain openshift.io
+function annotate_crd() {
+  script='/^metadata:/a\
+\ \ annotations:\
+\ \ \ \ exclude.release.openshift.io/internal-opernshift-hosted: "true"'
+  input="${1}"
+  output="${2}"
+  sed -e "${script}" "${input}" > "${output}"
+}
+
+go run ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen crd paths=./pkg/apis/...
 
 echo "Copying generated CRDs"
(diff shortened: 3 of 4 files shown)


enxebre:
hey @csrwng on the CRD I have same concern that https://github.com/openshift/machine-api-operator/pull/503#issuecomment-593838392

csrwng:
@enxebre for this one I updated the script and I can generate new CRDs. However, the generated CRDs are sufficiently different that I'd rather let the team decide when they want to regenerate them. For now I only manually inserted the annotations.

JoelSpeed:
Similar to https://github.com/openshift/machine-api-operator/pull/503#pullrequestreview-368689170, I wonder if the additional python dependency might break some CI generation somewhere?

csrwng:
Made the same change as in https://github.com/openshift/machine-api-operator/pull/503. Also ran the generate script, which seems like it would fix https://bugzilla.redhat.com/show_bug.cgi?id=1769004

install/01_clusterautoscaler.crd.yaml
```diff
@@ -1,159 +1,171 @@
...
-                  - min
-                  - max
-                  type: object
-                gpus:
-                  description: Minimum and maximum number of different GPUs in cluster,
-                    in the format <gpu_type>:<min>:<max>. Cluster autoscaler will
-                    not scale the cluster beyond these numbers. Can be passed multiple
-                    times.
-                  items:
+  scope: ""
```
JoelSpeed: I think we need to add some hints for the CRD generator, on both this and the machine autoscaler since the `scope` field has changed to an empty string.

It defaults to namespaced so this would break the `Cluster` scoped resource here

I think there needs to be a `// +kubebuilder:resource:scope=Cluster` tag added to the resource definition

csrwng: done

install/02_machineautoscaler.crd.yaml
```diff
@@ -1,131 +1,139 @@
...
-                  description: Name specifies a name of an object, e.g. worker-us-east-1a.
-                    Scalable resources are expected to exist under a single namespace.
-                  minLength: 1
-                  type: string
-              required:
-              - kind
-              - name
-              type: object
-          type: object
+  scope: ""
```
JoelSpeed: Non blocking nit: At some point we should add the resource here too, but since it's the default value I don't mind this being done later

install/01_clusterautoscaler.crd.yaml
```diff
@@ -1,159 +1,171 @@
...
-                  - min
-                  - max
-                  type: object
-                gpus:
-                  description: Minimum and maximum number of different GPUs in cluster,
-                    in the format <gpu_type>:<min>:<max>. Cluster autoscaler will
-                    not scale the cluster beyond these numbers. Can be passed multiple
-                    times.
-                  items:
+  version: v1
```
enxebre: where are all this changes to the CRD spec are coming from?

enxebre: cc @bison PTAL

csrwng: @enxebre This is the result of running crd generation with the more recent version of `controller-tools`. It looks like crd generation hadn't been run for a while on this repo.

csrwng: To make this a lower impact change, we could leave things as is (not run crd generation) and make those changes later in time, since the change is not required for what we need on the IBM work (all we need is the additional annotations)

enxebre: That sounds good. Let's add annotation manually but leave the script to generate in a follow up PR. So this can be back-ported minimizing impact for 4.4

csrwng: Thanks @enxebre, now changed to only add the annotations. I left the script change in there for when crds are generated in the future.

Issue #135, opened 2020-03-03
Add exclusion annotation in Prometheus Rules for the hosted use case


ryan-cradick:
The Prometheus Rules for this operator are being applied in the hosted use case. The exclude annotation needs to be added. This leads to an invalid alert.

https://github.com/openshift/cluster-autoscaler-operator/blob/master/install/09_alertrules.yaml

```
  annotations:
    exclude.release.openshift.io/internal-openshift-hosted: "true"
```

enxebre:
https://github.com/openshift/cluster-autoscaler-operator/pull/134/files

ryan-cradick:
Duplicate issue. Closing.

PR #136, opened 2020-03-03
Exclude PromethusRules for hosted use case

ryan-cradick:


ryan-cradick:
Issue: https://github.com/openshift/cluster-autoscaler-operator/issues/135

PR #149, opened 2020-04-27
Bug 1796412: cluster-reader is unable to view machine resources

Danil-Grigorev:
Added cluster-reader permissions to `cluster-autoscaler` cluster-role

### DIFF
--- a/install/10_cluster_reader_rbac.yaml
+++ b/install/10_cluster_reader_rbac.yaml
@@ -1,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cluster-autoscaler-operator:cluster-reader
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
+rules:
+- apiGroups:
+  - autoscaling.openshift.io
+  resources:
+  - '*'
+  verbs:
+  - get
+  - list
+  - watch


install/03_rbac.yaml
```diff
@@ -3,6 +3,8 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: cluster-autoscaler-operator
+  labels:
```
enxebre, line 6: this clusterRole has more permissions than the cluster reader should have.
We need to create one particular clusterRole reader for it, e.g https://github.com/openshift/cloud-credential-operator/blob/aa07b32fc3440d30bc90af636ca9ccd62e70c58a/manifests/00_clusterreader_clusterrole.yaml

hack/gen-crd.sh
```diff
@@ -3,12 +3,12 @@
 set -eu
 
 function annotate_crd() {
-  script='/^metadata:/a\
-\ \ annotations:\
+  script1='/^  annotations:/a\
 \ \ \ \ exclude.release.openshift.io/internal-openshift-hosted: "true"'
```
enxebre, line 7: this does not seem related to this PR, please let's drop it.

install/01_clusterautoscaler.crd.yaml
```diff
@@ -17,7 +17,6 @@ spec:
     - ca
     singular: clusterautoscaler
   scope: Cluster
-  preserveUnknownFields: false
```
alexander-demichev, line 20: Is this required?  @enxebre I remember you introduced `preserveUnknownFields: false` intentionally

install/02_machineautoscaler.crd.yaml
```diff
@@ -38,7 +38,6 @@ spec:
     - ma
     singular: machineautoscaler
   scope: Namespaced
-  preserveUnknownFields: false
```
enxebre, line 41: this does not seem related to this PR, please let's drop it.

install/01_clusterautoscaler.crd.yaml
alexander-demichev: I see the comment below now

hack/gen-crd.sh
Danil-Grigorev: @JoelSpeed going to include the autogen fix for crds in another pr, as I agree it is out of scope

install/10_cluster_reader_rbac.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cluster-autoscaler-operator:cluster-reader
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-cluster-reader: "true"
```
enxebre, line 6: is there any docs we can refer here for this label?

Danil-Grigorev: Under already created aggregation rule this ClusterRole will be combined with `cluster-reader` list of permissions, according to https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles

PR #151, opened 2020-05-20
pkg/operator/status: Change ReasonEmpty to ReasonAsExpected

wking:
We've used `ReasonEmpty` since 72cd9c4d76 (#23).  But if we feel like we have a message we want to set to help humans understand the condition, we should be setting a reason string for machines too.  AsExpected follows [the library-go precedent][1].

[1]: https://github.com/openshift/library-go/blob/94c59dec54be25c8527e51e8c0a885712aeb01b5/pkg/operator/status/condition.go#L67

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -22,7 +22,7 @@ import (
 
 // Reason messages used in status conditions.
 const (
-	ReasonEmpty             = ""
+	ReasonAsExpected        = "AsExpected"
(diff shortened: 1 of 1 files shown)


wking:
/assign @michaelgugino

Since this is basically the same change as the one you approved [here][1].

[1]: https://github.com/openshift/cluster-machine-approver/pull/72#pullrequestreview-417133149

PR #152, opened 2020-05-28
add metrics.md to document exposed metrics

elmiko:
This change adds a document to describe the metrics available from this
operator.

/kind documentation

### DIFF
--- a/docs/dev/metrics.md
+++ b/docs/dev/metrics.md
@@ -1,0 +1,89 @@
+# CAO Metrics
+
+The Cluster Autoscaler Operator(CAO) and the Cluster Autoscaler(CA) use the
+[Prometheus project](https://prometheus.io/) to expose metrics. To consume
+these metrics directly from the CAO or the CA you will need to perform
+HTTP GET requests to a specific port and URI on each application(CAO/CA). The
+URI for all metrics is `/metrics`, see the Prometheus documentation for query
+parameter options. To find the exposed metrics port for the CAO you can either
+inspect the Deployment resource or the
+[install manifest](https://github.com/openshift/cluster-autoscaler-operator/blob/master/install/07_deployment.yaml)
+to find the environment variable `METRICS_PORT`, the default value for this is `9191`.
+Finding the metrics port for the CA is slightly more complex, at the time of this
+writing there is no explicitly mentioned
+service for metrics on the CA so you should use the default port of `8085`.
+
+**Example CAO metrics scrape procedure**
+1. Forward the metrics port from the CAO to a local port
+   ```
+   $ oc port-forward -n openshift-machine-api deployment/cluster-autoscaler-operator 9191:9191
+   ```
+2. Perform an HTTP GET request on the local port
+   ```
+   $ curl http://localhost:9191/metrics
+   ```
+
+**Example CA metrics scrape procedure**
+1. Forward the metrics port from the CA to a local port
+   ```
+   $ oc port-forward -n openshift-machine-api deployment/cluster-autoscaler-default 8085:8085
+   ```
+2. Perform an HTTP GET request on the local port
+   ```
+   $ curl http://localhost:8085/metrics
+   ```
+
+The Cluster Autoscaler Operator reports the following metrics:
+
+## Metrics provided by the controller runtime
+
+The [controller runtime](https://github.com/kubernetes-sigs/controller-runtime)
+integration with the operator provides several metrics about the internals of the
+operator. You can find more information about these metrics names and their labels through
+the following links:
+
+### Kubernetes controller metrics
+
+These metric names begin with `controller_runtime_reconcile_`. The labels
+`controller="cluster_autoscaler_controller"` and
+`controller="machine_autoscaler_controller"` can be used to refine queries against these metrics.
+* [Controller runtime reconciliation metrics implementation](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/metrics/metrics.go)
+
+### Admission webhook metrics
+
+These metric names begin with `controller_runtime_webhook_`.  The label
+`webhook="/validate-clusterautoscalers"` can be used to refine the queries for these metrics.
+* [Controller runtime webhook metrics implementation](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/internal/metrics/metrics.go)
+
+### API REST server metrics
+
+These metric names begin with `rest_client_` and `reflector_`. The labels
+`url` and `verb` can be used to refine these metrics further. The `url` label
+should contain the address to the Kubernetes API server, while `verb`
+containts the type of HTTP request (eg `"GET"`).
+* [Controller runtime Prometheus REST server metrics implementation](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/metrics/client_go_adapter.go)
+
+### Prometheus work queue metrics
+
+These metric names begin with `workqueue_`. The labels
+`name="cluster_autoscaler_controller"` and `name="machine_autoscaler_controller"`
+can be used to refine queries against these metrics.
+* [Controller runtime Prometheus work queue metrics implementation](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/metrics/workqueue.go)
+
+## Metrics about the Prometheus collectors
+
+Prometheus provides some default metrics about the internal state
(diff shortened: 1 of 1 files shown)


enxebre:
Can we include a ref to the metrics related to the cluster autoscaler https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/controller/clusterautoscaler/monitoring.go
https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/proposals/metrics.md?

elmiko:
i'm digging deeper into some of these metrics sources in hopes that we can provide links back to the original sources.

docs/dev/metrics.md
```diff
@@ -0,0 +1,53 @@
...
+### Admission webhook metrics
+
+The label `webhook="/validate-clusterautoscalers"` can be used to refine the
+queries for these metrics.
+* [Controller runtime webhook metrics implementation](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/webhook/internal/metrics/metrics.go)
+
+### Prometheus REST server metrics
+
+The `url` label can be quite useful for querying these metrics, here are a few
+of the available URL values to use: `"https://172.30.0.1:443/%7Bprefix%7D"`, `"https://172.30.0.1:443/apis?timeout=32s"`.
```
michaelgugino, line 27: From where should I utilize these URLs?  I'm assuming these IPs are subject to change?

elmiko: these are the in-pod IP addresses that are assigned to the prometheus http server, they are used as labels during promql calls to isolate the different endpoints. in my tests it looked like these IP addresses always came up the same, but i have to imagine it would change if the address range changed for the container network.

JoelSpeed: Just as a heads up, I believe `172.30.0.1` is the service IP of the Kubernetes service in the default namespace on an OpenShift cluster, ie, it's the service that fronts the API server, not sure why this is coming up WRT prometheus though

elmiko: is it possible we are just seeing a local ip address that only exists within the pod network for these containers?

JoelSpeed: The pod network on openshift is a `10.x` subnet, the `172.30.x` is used for services, perhaps the metrics are for requests that CAO is making to the Kubernetes API?

elmiko: could be, i'll take some time to dig in a little and see if i can learn more here.

i guess another way to look at this is, should we just remove this info and point to the docs in the prometheus project? my concern is that we get focused on this metric when it seems like a minor thing.

JoelSpeed: This is a valid concern, I agree, maybe we just drop that little bit

elmiko: first, thank you both for pushing a little on this one it's been a fascinating investigation. as near as i can tell, these metrics are being generated by client-go using the controller-runtime package. i'm pretty sure [this is where it starts](https://github.com/kubernetes/client-go/blob/master/rest/request.go#L720). it appears to be showing all the outbound requests that are being made by the client-go inside the operator. which would square up with what Joel is saying about the IP address being the root API server.

so, maybe explain that bit here and give the example IP address? or is that too much detail?

elmiko:
updated to add more information about the controller-runtime based metrics. also removed the specific urls for the api server, with a description instead.

JoelSpeed:
LGTM, I'll leave the label for @michaelgugino in case he has further comments

docs/dev/metrics.md
```diff
@@ -0,0 +1,56 @@
...
+### Prometheus work queue metrics
+
+These metric names begin with `workqueue_`. The labels
+`name="cluster_autoscaler_controller"` and `name="machine_autoscaler_controller"`
+can be used to refine queries against these metrics.
+* [Controller runtime Prometheus work queue metrics implementation](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/metrics/workqueue.go)
+
+## Metrics about the Prometheus collectors
+
+Prometheus provides some default metrics about the internal state
```
michaelgugino, line 42: I'm not clear on what this section means.  Are these metrics collected automatically by virtue of using some prometheus library in the sig-controller-runtime code?

docs/dev/metrics.md
```diff
@@ -0,0 +1,56 @@
...
+of the running process and the metric collection. You can find more information
+about these metric names and their labels through the following links:
+
+* [Prometheus documentation, Standard and runtime collectors](https://prometheus.io/docs/instrumenting/writing_clientlibs/#standard-and-runtime-collectors)
+* [Prometheus client Go language collectors](https://github.com/prometheus/client_golang/blob/master/prometheus/go_collector.go)
+
+# Cluster Autoscaler Metrics
```
michaelgugino, line 49: Nit:  This seems like it could replace the first H1 at the top of this doc.

docs/dev/metrics.md
michaelgugino: Okay, yeah, that's how it works.  Wasn't clear what this was about.

How about an example.
```process_cpu_seconds_total{job="cluster-autoscaler-operator"}```

File: docs/dev/metrics.md [hunk 4]
```diff
@@ -0,0 +1,56 @@
...
+### Kubernetes controller metrics
+
+These metric names begin with `controller_runtime_reconcile_`. The labels
+`controller="cluster_autoscaler_controller"` and
+`controller="machine_autoscaler_controller"` can be used to refine queries against these metrics.
+* [Controller runtime reconciliation metrics implementation](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/internal/controller/metrics/metrics.go)
+
+### Admission webhook metrics
+
+These metric names begin with `controller_runtime_webhook_`.  The label
```
michaelgugino, line 21: Are these actually being scraped?  I did some queries on 4.5, and didn't get anything back for "controller_runtime_webhook_requests_total" and other metrics in this link.

docs/dev/metrics.md
```diff
@@ -0,0 +1,56 @@
...
+## Metrics provided by the controller runtime
+
+The [controller runtime](https://github.com/kubernetes-sigs/controller-runtime)
+integration with the operator provides several metrics about the internals of the
+operator. You can find more information about these metrics names and their labels through
+the following links:
+
+### Kubernetes controller metrics
+
+These metric names begin with `controller_runtime_reconcile_`. The labels
```
michaelgugino, line 14: Are these metrics actually getting collected?  No CAO/MAO stuff being reported in 4.5.

docs/dev/metrics.md
elmiko: this one is specifically about the autoscaler though, the top level link is about the operator. i was trying to draw a distinction, maybe there is a better way to specify?

docs/dev/metrics.md
elmiko: sure, i can add an example. i do  wonder if i shouldn't update the mao doc as well then since it has the same clause in it?

https://github.com/openshift/machine-api-operator/blob/master/docs/dev/metrics.md#metrics-about-the-prometheus-collectors

note that the sample metric in that section is specifically added by the mao, that isn't the case here.

docs/dev/metrics.md
elmiko: define "queries".

these metrics may not get collected by telemetry, but they are exposed from the operator. i was able to scrape them from a running cao on 4.5.

docs/dev/metrics.md
elmiko: they may not be collected, but they are exposed by the metrics endpoint for both operators.

docs/dev/metrics.md
michaelgugino: Yeah, they're not getting collected by prometheus inside the cluster, I'm not talking about telemetry.  They may be exposed but not collected, as you say.

docs/dev/metrics.md
michaelgugino: Same here.  Not getting collected by prometheus.  I think some other operators do have this information being collected in prometheus.  In any case, we should specify that we are/aren't collecting these?

elmiko: >  In any case, we should specify that we are/aren't collecting these?

that's a good question, my gut feeling is that we should document all the possible metrics someone could scrape from the operator. given that the collection/telemetry stuff is controlled from a different location, perhaps we should drop a link to that repo so readers can evaluate what is being exported.

PR #156, opened 2020-07-08
Bug 1854907: Fix #155 - ensure full boolean arguments

### LINKED ISSUE #155: Config logic for skip-nodes-with-local-storage is flawed
There is an example to set the following option:

https://github.com/openshift/cluster-autoscaler-operator/blob/9c4a47c56f7fb7cd2677d4adbda48f9a84d3be6a/examples/clusterautoscaler.yaml#L9

However, when setting this option to `false` nothing happens. The deployment is not getting updated.

This is because the configuration logic is flawed:

https://github.com/openshift/cluster-autoscaler-operator/blob/9c4a47c56f7fb7cd2677d4adbda48f9a84d3be6a/pkg/controller/clusterautoscaler/clusterautoscaler.go#L95-L97

But you want the autoscaler to run with `--skip-nodes-with-local-storage=false` if you want to scale down nodes with pods using emptyDir.

duritong:
Currently we did not place boolean arguments if they were not true,
which didn't allow to set an option to false.

While some arguments might be true by default (SkipNodesWithLocalStorage)
others (balanceSimilarNodeGroups) are false by default. So let's
set the value explicitely for the few boolean values.

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler.go
@@ -84,16 +84,16 @@ func AutoscalerArgs(ca *v1.ClusterAutoscaler, cfg *Config) []string {
 		args = append(args, ScaleDownArgs(s.ScaleDown)...)
 	}
 
-	if ca.Spec.BalanceSimilarNodeGroups != nil && *ca.Spec.BalanceSimilarNodeGroups {
-		args = append(args, BalanceSimilarNodeGroupsArg.String())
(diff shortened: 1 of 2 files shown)


duritong:
Can you please backport it to release-4.4 and following.

elmiko:
i tested this against a 4.6 nightly, everything worked as expected. thanks @duritong !
/approve

enxebre:
/retitle "Bug 1854907: Fix #155 - ensure full boolean arguments"

PR #158, opened 2020-07-31
Bug 1861642: Add maxNodeProvisionTime for baremetal

hardys:
In baremetal environments it can easily take longer than the CA
default of 15mins for the node to become active after a scale-out
action.

The CA supports --max-node-provision-time[1] so adding support to
enable configuration of that value should allow tuning of the
time such that it's more suited to baremetal.

[1] https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca

### DIFF
--- a/install/01_clusterautoscaler.crd.yaml
+++ b/install/01_clusterautoscaler.crd.yaml
@@ -49,6 +49,14 @@ spec:
                   feature flag. Should CA ignore DaemonSet pods when calculating resource
                   utilization for scaling down. false by default
                 type: boolean
+              maxNodeProvisionTime:
+                description: Maximum time CA waits for node to be provisioned.
+                             Expects an unsigned duration string of decimal
+                             numbers each with optional fraction and a unit
+                             suffix, eg "300ms", "1.5h" or "2h45m". Valid time
+                             units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+                pattern: ^([0-9]+(\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$
+                type: string
               maxPodGracePeriod:
                 description: Gives pods graceful termination time before scaling down
                 format: int32
--- a/pkg/apis/autoscaling/v1/clusterautoscaler_types.go
+++ b/pkg/apis/autoscaling/v1/clusterautoscaler_types.go
@@ -19,6 +19,10 @@ type ClusterAutoscalerSpec struct {
 	// Gives pods graceful termination time before scaling down
 	MaxPodGracePeriod *int32 `json:"maxPodGracePeriod,omitempty"`
 
+	// Maximum time CA waits for node to be provisioned
+	// +kubebuilder:validation:Pattern=([0-9]*(\.[0-9]*)?[a-z]+)+
+	MaxNodeProvisionTime string `json:"maxNodeProvisionTime,omitempty"`
+
 	// To allow users to schedule "best-effort" pods, which shouldn't trigger
 	// Cluster Autoscaler actions, but only run when there are spare resources available,
 	// More info: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-does-cluster-autoscaler-work-with-pod-priority-and-preemption
--- a/pkg/apis/autoscaling/v1/clusterautoscaler_types.go
+++ b/pkg/apis/autoscaling/v1/clusterautoscaler_types.go
@@ -20,7 +20,7 @@ type ClusterAutoscalerSpec struct {
 	MaxPodGracePeriod *int32 `json:"maxPodGracePeriod,omitempty"`
 
 	// Maximum time CA waits for node to be provisioned
-	// +kubebuilder:validation:Pattern=([0-9]*(\.[0-9]*)?[a-z]+)+
+	// +kubebuilder:validation:Pattern=^([0-9]+(\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$
 	MaxNodeProvisionTime string `json:"maxNodeProvisionTime,omitempty"`
 
 	// To allow users to schedule "best-effort" pods, which shouldn't trigger
--- a/pkg/controller/clusterautoscaler/clusterautoscaler.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler.go
@@ -44,6 +44,7 @@ const (
 	ScaleDownDelayAfterFailureArg   AutoscalerArg = "--scale-down-delay-after-failure"
 	ScaleDownUnneededTimeArg        AutoscalerArg = "--scale-down-unneeded-time"
 	MaxNodesTotalArg                AutoscalerArg = "--max-nodes-total"
+	MaxNodeProvisionTimeArg         AutoscalerArg = "--max-node-provision-time"
 	CoresTotalArg                   AutoscalerArg = "--cores-total"
 	MemoryTotalArg                  AutoscalerArg = "--memory-total"
 	GPUTotalArg                     AutoscalerArg = "--gpu-total"
@@ -68,6 +69,11 @@ func AutoscalerArgs(ca *v1.ClusterAutoscaler, cfg *Config) []string {
 
 	if ca.Spec.MaxPodGracePeriod != nil {
 		v := MaxGracefulTerminationSecArg.Value(*s.MaxPodGracePeriod)
+		args = append(args, v)
+	}
+
+	if ca.Spec.MaxNodeProvisionTime != "" {
+		v := MaxNodeProvisionTimeArg.Value(s.MaxNodeProvisionTime)
 		args = append(args, v)
 	}
 
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_test.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_test.go
@@ -34,6 +34,7 @@ const (
 var (
 	ScaleDownUnneededTime        = "10s"
 	ScaleDownDelayAfterAdd       = "60s"
+	MaxNodeProvisionTime         = "30m"
 	PodPriorityThreshold   int32 = -10
 	MaxPodGracePeriod      int32 = 60
 	MaxNodesTotal          int32 = 100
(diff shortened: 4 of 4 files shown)


hardys:
I've tested this locally to confirm no regressions, but I can't see a way to get the new CRD definition - I guess the CVO has an old copy, any tips on how to work around that for local testing?

install/01_clusterautoscaler.crd.yaml
```diff
@@ -49,6 +51,10 @@ spec:
                   feature flag. Should CA ignore DaemonSet pods when calculating resource
                   utilization for scaling down. false by default
                 type: boolean
+              maxNodeProvisionTime:
+                description: Maximum time CA waits for node to be provisioned
+                pattern: ([0-9]*(\.[0-9]*)?[a-z]+)+
```
elmiko, line 56: just a question about the regex, do we need to restrict the alphanumeric on this? it seems like it would accept a wide range of inputs.

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
```diff
@@ -145,7 +146,8 @@ func TestAutoscalerArgs(t *testing.T) {
 
 	expectedMissing := []string{
 		"--scale-down-delay-after-delete",
-		"--scale-down-delay-after-failure",
```
elmiko, line 148: i'm not understanding why this line is being removed

install/01_clusterautoscaler.crd.yaml
hardys: Thanks for the prompt review @elmiko :+1: 

Re the regex, I copied it from the similar interfaces in the scaleDown section e.g delayAfterAdd etc - those support multiple formats e.g "10m" "30s" etc - from the CA docs it looks like --max-node-provision-time supports the same (they're all flag.Duration):

https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/main.go#L137

Looking at the underlying code, it uses `time.ParseDuration` which does describe the exact same regex:

https://golang.org/src/time/format.go?s=40541:40587#L1364

elmiko: cool, works for me!

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
hardys: No this looks like a mistake, thanks - will fix!

elmiko:
@enxebre @JoelSpeed ptal

pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -71,6 +72,11 @@ func AutoscalerArgs(ca *v1.ClusterAutoscaler, cfg *Config) []string {
 		args = append(args, v)
 	}
 
+	if ca.Spec.MaxNodeProvisionTime != nil {
+		v := MaxNodeProvisionTimeArg.Value(*s.MaxNodeProvisionTime)
+		args = append(args, v)
+	}
```
JoelSpeed, line 78: Will this break the autoscaler if `*ca.Spec.MaxNodeProvisionTime == ""`? Is there actually any benefit to the field being a string pointer?

install/01_clusterautoscaler.crd.yaml
JoelSpeed: We use an alternative regex already for the MHC `time.Duration` fields that is more restrictive than your current one and should guarantee a valid duration, perhaps we could use that one instead? https://github.com/openshift/machine-api-operator/blob/27db623eabbb97e5cec9a47863da40d119b250e8/install/0000_30_machine-api-operator_07_machinehealthcheck.crd.yaml#L76

install/01_clusterautoscaler.crd.yaml
```diff
@@ -5,6 +5,8 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     exclude.release.openshift.io/internal-openshift-hosted: "true"
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.2.9-0.20200331153640-3c5446d407dd
```
JoelSpeed, line 9: Can you remove these two lines please, I realise they're generated by they are unnecessary and will actually break the yaml by defining `annotations` twice?

install/02_machineautoscaler.crd.yaml
```diff
@@ -5,6 +5,8 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     exclude.release.openshift.io/internal-openshift-hosted: "true"
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.2.9-0.20200331153640-3c5446d407dd
```
JoelSpeed, line 9: Can you remove these two lines please, I realise they're generated by they are unnecessary and will actually break the yaml by defining `annotations` twice

pkg/controller/clusterautoscaler/clusterautoscaler.go
hardys: This is consistent with most other fields in the spec - I think they are defined as pointers deliberately, or the `omitempty` wouldn't work - golang would give a default value for each field vs a nil pointer.

install/01_clusterautoscaler.crd.yaml
hardys: Done!

install/01_clusterautoscaler.crd.yaml
hardys: Done - and thanks for pointing to nodeStartupTimeout - I suspect that may also need adjustment for this to work well in a baremetal env due to the time taken for reboots.

pkg/controller/clusterautoscaler/clusterautoscaler.go
hardys: https://www.sohamkamani.com/golang/2018-07-19-golang-omitempty/ has a nice explanation on this - AFAICS the user can't pass `""` because that's treated as a nil interface value when marshalling the yaml and specifying `omitempty`

JoelSpeed: I don't think that's the case for string values, if `""` is not a value we are allowing, then because that is the zero value, we can use that as the unset case

See this playground example: https://play.golang.org/p/YaETDGcJsX3, which demonstrates that with a string pointer, you can end up with a pointer to an empty string, which would then get passed through and result in the empty string value, this example also demonstrates that the empty string reference marshalling for omitempty works as well, see on both examples how the field isn't marshalled when printed out

JoelSpeed: In each of the other fields in this spec, there is a clear distinction in meaning between `nil` and the zero value, eg, `*bool` is commonly used when you want to allow a boolean to default to true, by seeing that it is nil, you can default it, otherwise you have no way to no if the zero value (`false`) is there because the user set it, or if it's just because it's the zero value

hardys: @JoelSpeed Ah I see now, thanks for clarifying.

I think the source of my confusion is that in ScaleDownConfig the values all appear to have the same problem - the interface is similar so I just copied the existing pattern:

```
$ grep "\*string"  pkg/apis/autoscaling/v1/clusterautoscaler_types.go 
	MaxNodeProvisionTime *string `json:"maxNodeProvisionTime,omitempty"`
	DelayAfterAdd *string `json:"delayAfterAdd,omitempty"`
	DelayAfterDelete *string `json:"delayAfterDelete,omitempty"`
	DelayAfterFailure *string `json:"delayAfterFailure,omitempty"`
	UnneededTime *string `json:"unneededTime,omitempty"`

$ grep -e DelayAfter -e UnneededTime  pkg/controller/clusterautoscaler/clusterautoscaler.go | grep nil
	if sd.DelayAfterAdd != nil {
	if sd.DelayAfterDelete != nil {
	if sd.DelayAfterFailure != nil {
	if sd.UnneededTime != nil {
```

If you agree these should be modified in a similar way I can push a followup (and the regex could be changed to be consistent with what's been discussed here?).

install/01_clusterautoscaler.crd.yaml
```diff
@@ -49,6 +49,14 @@ spec:
                   feature flag. Should CA ignore DaemonSet pods when calculating resource
                   utilization for scaling down. false by default
                 type: boolean
+              maxNodeProvisionTime:
+                description: Maximum time CA waits for node to be provisioned.
+                             Expects an unsigned duration string of decimal
+                             numbers each with optional fraction and a unit
+                             suffix, eg "300ms", "1.5h" or "2h45m". Valid time
+                             units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+                pattern: ^([0-9]+(\.[0-9]+)?(ns|us|µs|ms|s|m|h)*)+$
```
JoelSpeed, line 58: I found a bug in this format recently, units are actually required otherwise the duration won't parse, this is the fix
```suggestion
                pattern: ^([0-9]+(\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$
```

pkg/controller/clusterautoscaler/clusterautoscaler.go
JoelSpeed: Yeah I think we should probably look into those other ones as well, as they are string pointers, there's a possibility we could have a bug where a user enters the empty string accidentally and we add a flag such as `--unneededTime= ` which would cause the autoscaler to throw a flag parse error on startup

JoelSpeed:
I'm happy to approve this once the format is updated, apologies, literally found out there's a small bug in it earlier today 😅

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -19,6 +19,10 @@ type ClusterAutoscalerSpec struct {
 	// Gives pods graceful termination time before scaling down
 	MaxPodGracePeriod *int32 `json:"maxPodGracePeriod,omitempty"`
 
+	// Maximum time CA waits for node to be provisioned
+	// +kubebuilder:validation:Pattern=([0-9]*(\.[0-9]*)?[a-z]+)+
```
JoelSpeed, line 23: This one needs to match as well else you will overwrite the pattern when we run `make gen-crd` next
```suggestion
	// +kubebuilder:validation:Pattern=^([0-9]+(\.[0-9]+)?(ns|us|µs|ms|s|m|h))+$
```

hardys: Ah yeah I missed that - thanks!

hardys:
@elmiko - please could you revisit this when you get a moment - thanks! :)

Issue #153, opened 2020-06-24
Document alerts generated by the cluster autoscaler


elmiko:
There are several alerts that can be generated by the cluster autoscaler, we should create a document to highlight them and give guidance on user actions where possible. [This PR from machine-api-operator](https://github.com/openshift/machine-api-operator/pull/606) gives an example of how we can start crafting this document and where to place it.

PR #163, opened 2020-09-03
Bug 1819029: add a document about autoscaler alerts

elmiko:
This change adds the alerts.md file to the user documentation, it
describes in detail the alerts that can be generated, their causes, and
suggested solutions.


elmiko:
@michaelgugino would love your feedback on this

elmiko:
updated to make the faq link point to the top of the doc

docs/user/alerts.md
```diff
@@ -0,0 +1,75 @@
...
+unschedulable pods because of this condition.
+
+### Query
+```
+# for: 20m
+cluster_autoscaler_unschedulable_pods_count{service="cluster-autoscaler-default"} > 0
+```
+
+### Possible Causes
+* The cluster autoscaler does not have enough resources to increase the size of
```
michaelgugino, line 18: How about

"The autoscaler is unable to create new machines due to resource limits on MachineAutoscalers"

docs/user/alerts.md
```diff
@@ -0,0 +1,75 @@
...
+  them.
+
+### Resolution
+In many cases this alert is normal and expected depending on the configuration
+of the autoscaler. If you believe this alert is describing abnormal behavior,
+check the configuration of your ClusterAutoscaler resource to ensure that you
+have specified a large enough pool of nodes, with sufficient memory and CPU core
+values to cover the entire cluster.
+
+In rare cases it is possible that the cloud provider is taking longer than 20
```
michaelgugino, line 30: We need to adjust this alert to longer than 20 minutes in the future.

docs/user/alerts.md
elmiko: this makes me think we need 2 entries here, the one you are proposing and one about the ClusterAutoscaler resources. i think you can hit this alert if your MachineAutoscaler doesn't have enough replicas or if your ClusterAutoscaler doesn't define enough nodes/cpu/ram. wdyt?

michaelgugino: I concur.

elmiko:
updated the causes and resolution for ClusterAutoscalerUnschedulablePods

elmiko:
fixes #153

JoelSpeed (commented):
Looks good, one nit, typically in docs I try to capitalise Kubernetes resource names (Pods, Machines, MachineAutoscalers, etc) to make it obvious that these are objects that we are referring to rather than generic concepts (is a "machine" referring to a Machine or a virtual machine?), not sure if others care about this as much so feel free to ignore

 /approve

elmiko:
> Looks good, one nit, typically in docs I try to capitalise Kubernetes resource names (Pods, Machines, MachineAutoscalers, etc) to make it obvious that these are objects that we are referring to rather than generic concepts (is a "machine" referring to a Machine or a virtual machine?), not sure if others care about this as much so feel free to ignore

it's funny you mention this. i usually capitalize all the kube resource names as well, but it started to get oddly pedantic at some point so i reverted it. i'm happy to put them back though, this might be something we should look to the kube/openshift docs for inspiration (eg maybe we should enforce ourselves to always do it)

PR #164, opened 2020-09-15
Bug 1879162: ensure full boolean arguments

elmiko:
Currently we did not place boolean arguments if they were not true,
which didn't allow to set an option to false.

While some arguments might be true by default (SkipNodesWithLocalStorage)
others (balanceSimilarNodeGroups) are false by default. So let's
set the value explicitely for the few boolean values.

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_test.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_test.go
@@ -169,9 +169,32 @@ func TestAutoscalerArgEnabled(t *testing.T) {
 	args := AutoscalerArgs(ca, &Config{CloudProvider: TestCloudProvider, Namespace: TestNamespace})
 
 	expected := []string{
-		fmt.Sprintf("--balance-similar-node-groups"),
-		fmt.Sprintf("--ignore-daemonsets-utilization"),
-		fmt.Sprintf("--skip-nodes-with-local-storage"),
+		"--balance-similar-node-groups=true",
+		"--ignore-daemonsets-utilization=true",
+		"--skip-nodes-with-local-storage=true",
+	}
+
+	for _, e := range expected {
+		if !includeString(args, e) {
+			t.Fatalf("missing arg: %s", e)
+		}
+	}
+}
+
+// TestAutoscalerArgEnabledWithFalse validates that args appear in the autoscaler args when
+// set to false in the ClusterAutoscalerSpec.
+func TestAutoscalerArgEnabledWithFalse(t *testing.T) {
+	ca := NewClusterAutoscaler()
+	ca.Spec.BalanceSimilarNodeGroups = pointer.BoolPtr(false)
+	ca.Spec.IgnoreDaemonsetsUtilization = pointer.BoolPtr(false)
+	ca.Spec.SkipNodesWithLocalStorage = pointer.BoolPtr(false)
+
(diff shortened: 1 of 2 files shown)


elmiko:
@duritong ptal

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
```diff
@@ -169,9 +169,32 @@ func TestAutoscalerArgEnabled(t *testing.T) {
 	args := AutoscalerArgs(ca, &Config{CloudProvider: TestCloudProvider, Namespace: TestNamespace})
 
 	expected := []string{
-		fmt.Sprintf("--balance-similar-node-groups"),
-		fmt.Sprintf("--ignore-daemonsets-utilization"),
-		fmt.Sprintf("--skip-nodes-with-local-storage"),
+		fmt.Sprintf("--balance-similar-node-groups=true"),
+		fmt.Sprintf("--ignore-daemonsets-utilization=true"),
+		fmt.Sprintf("--skip-nodes-with-local-storage=true"),
```
JoelSpeed, line 174: I know this is copying an existing pattern, but, why are we fmt printing strings with no format characters in them? Shouldn't we just use the strings raw?
```suggestion
		"--balance-similar-node-groups=true",
		"--ignore-daemonsets-utilization=true",
		"--skip-nodes-with-local-storage=true",
```

JoelSpeed (commented):
/approve

One nit for the tests that we could clean up while touching this code

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
elmiko: i can change this up

elmiko:
updated to remove the unneeded `fmt.Sprintf` in the unit tests.

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
```diff
@@ -169,9 +169,32 @@ func TestAutoscalerArgEnabled(t *testing.T) {
...
+	ca.Spec.BalanceSimilarNodeGroups = pointer.BoolPtr(false)
+	ca.Spec.IgnoreDaemonsetsUtilization = pointer.BoolPtr(false)
+	ca.Spec.SkipNodesWithLocalStorage = pointer.BoolPtr(false)
+
+	args := AutoscalerArgs(ca, &Config{CloudProvider: TestCloudProvider, Namespace: TestNamespace})
+
+	expected := []string{
+		fmt.Sprintf("--balance-similar-node-groups=false"),
+		fmt.Sprintf("--ignore-daemonsets-utilization=false"),
+		fmt.Sprintf("--skip-nodes-with-local-storage=false"),
```
JoelSpeed, line 197: Can we also remove the `fmt` here please

elmiko: sure thing

elmiko:
updated _other_ set of unnecessary `fmt.Sprintf` calls

sunzhaohua2:
/bugzilla cc-qa
(I'm following the [DPTP-660](https://issues.redhat.com/browse/DPTP-660) to do the pre-merge verifying)

sunzhaohua2:
Already pre-merge verified in [1879162#c1](https://bugzilla.redhat.com/show_bug.cgi?id=1879162#c1), so per DPTP-660 adding:
/lgtm

PR #171, opened 2020-10-07
Create clientsets for autoscaling API groups

JoelSpeed:
This PR adds generated clientsets for the autoscaling v1 and v1beta1 API groups.

This is all pretty much generated or copied from the MAO repository.


elmiko (approved):
thanks Joel, this makes sense to me.

i still have some concerns about maintenance and upkeep since we are generating these for external users, i would love to see us promoting more controller-runtime/dynamic client usage.

that said, i think we shouldn't block the wider usage of these types.
/lgtm

JoelSpeed:
Interest in this died down, closing until we get a refreshed interest

PR #172, opened 2020-10-08
add bugzilla component to OWNERS

elmiko:
This change adds the bugzilla component name to the OWNERS file to
comply with the prodsec requirements for OpenShift builds through ART.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,3 +1,5 @@
+component: "Cloud Compute"
+subcomponent: "Other Providers"
 approvers:
   - enxebre
   - frobware


JoelSpeed:
Should we also be including the subcomponent on this?

elmiko:
for this one i think we should not include the subcomponent. subcomponent is optional and the cao could be related in bugzilla to other providers as well (eg ovirt, openstack).

JoelSpeed:
Technically I think we should have a separate component for CAO, but I would expect any bug that is for the CAO (not the CA) to be related to the cluster-infrastructure team, not any provider team in particular

elmiko:
i tend to agree about having a separate component, or maybe just a separate subcomponent. imo, considering that "subcomponent" in the OWNERS is optional, we should opt to not use it to ensure we aren't over-scoping.

michaelgugino (changes requested):
/lgtm cancel

michaelgugino:
I don't think the subcomponent is actually optional.  BZ usually complains when I try to create a bug without setting it.

michaelgugino (approved):
/lgtm

PR #179, opened 2020-11-02
add current profile annotations to CVO manifests

guillaumerose:
This is matches openshift/enhancements#510 and doesn't change existing behavior

---

It was one of the first repos I changed and I missed the fact that some files can contain multiples manifests.


enxebre:
/lgtm
PTAL @JoelSpeed

PR #185, opened 2020-11-13
Bug 1879777: Remove namespace manifest

JoelSpeed:
This manifest is a duplicate of https://github.com/openshift/machine-api-operator/blob/master/install/0000_30_machine-api-operator_00_namespace.yaml

MAO runs at run level 30, whereas this component runs at run level 50.

The manifests differ presently only by an annotation (present in the MAO copy), but could end up drifting in the future.

To prevent CVO potentially flip-flopping on conflicting configs, we should have an authoritative (and single) copy of this manifest within the release bundle, this should be the one in MAO, hence we can remove this one.


Danil-Grigorev:
/approve
/lgtm

It is a simple fix.

JoelSpeed:
/retest

I had a look at the test failures, for some reason the Kube controller manager suddenly became unauthorized, very weird

alexander-demichev (approved):
/approve

PR #188, opened 2020-12-03
Bug 1915473: Annotate manifests for single-node-developer cluster profile

cfergeau:
This partially implements phase 1 of https://github.com/openshift/enhancements#482
and does not change behavior. Initially, all cluster-autoscaler-operator
manifests are included in the single-node-developer cluster profile.
Follow-on PRs may exclude any of these that are not needed in the
profile.

### DIFF
--- a/hack/gen-crd.sh
+++ b/hack/gen-crd.sh
@@ -5,7 +5,8 @@ set -eu
 function annotate_crd() {
   script1='/^  annotations:/a\
 \ \ \ \ exclude.release.openshift.io/internal-openshift-hosted: "true"\
-\ \ \ \ include.release.openshift.io/self-managed-high-availability: "true"'
+\ \ \ \ include.release.openshift.io/self-managed-high-availability: "true"\
+\ \ \ \ include.release.openshift.io/single-node-developer: "true"'
   script2='/^    controller-gen.kubebuilder.io\/version: .*$/d'
   input="${1}"
   output="${2}"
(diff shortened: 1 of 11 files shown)


elmiko:
hey, i just took a quick look at this and it looks ok to me but i want to read up on the enhancement a little more.

elmiko (changes requested):
after reading through the enhancement, i do not think we should be including the autoscaler components in the single node profile.

firstly, given the goals described in that doc the autoscaler should not be used with the single node profile.

secondly, the machine-api-operator is called out specifically as not being included and the autoscaler would require the machine-api to operate..

i think we should mark this component as false, or exclude it from inclusion in the single node profile. it wasn't clear to me from reading the enhancement if we should use the `include.release.openshift.io/single-node-developer: "false"` annotation at this time or if we should just not add it to this component.

cfergeau:
For now, we are including everything in the single-node-developer profile, as before disabling components, we need to make sure nothing requires it during the installation. But I agree that eventually it's likely we'll disable this.
See https://github.com/openshift/machine-api-operator/pull/774#issuecomment-740527434 for the corresponding machine-api discussion.

elmiko:
thanks for the clarification @cfergeau
/approve

cfergeau:
> thanks for the clarification @cfergeau
> /approve

Can you also add a /lgtm? Or is this going to be done by a second reviewer?

PR #190, opened 2020-12-18
Bug 1909108: Bump K8s dependencies to 1.20

JoelSpeed:
What it says on the tin. Bumps K8s dependencies to v1.20 to be inline with OCP 4.7


elmiko (approved):
/lgtm

PR #195, opened 2021-03-17
add ClusterAutoscalerExcessiveUnregisteredNodeCleanup alert

elmiko:
This change adds an alert to the cluster autoscaler for when it is
removing unregistered nodes that did not join the cluster.

Also updates the alert doc with new information.


elmiko:
this will need to wait for https://github.com/openshift/kubernetes-autoscaler/pull/196
/hold

docs/user/alerts.md
```diff
@@ -60,6 +61,28 @@ the Machine API resources to discover the root cause. For more information on
...
+rate(cluster_autoscaler_old_unregistered_nodes_removed_count[6h]) > 1
+```
+
+### Possible Causes
+* MachineSet, or cloud credential, configuration is causing machines to fail.
+* Network disruption is preventing machines from becoming nodes.
+
+### Resolution
+* Look at the cluster autoscaler events, or logs, to determine which Machines have been removed
+  and to which MachineSets they belong.
```
wking, line 80: Can we make the MachineSet name a label on the metric, and then include that name in the alert to make running this down a bit more convenient for folks who are triaging it?

elmiko: i think that's a great idea, but upstream did not like having a label with unbounded cardinality on the metric. i don't think it would be an issue in production, but i can understand their point so we decided to drop the label in favor of having the metric.

docs/user/alerts.md
```diff
@@ -60,6 +61,28 @@ the Machine API resources to discover the root cause. For more information on
...
+* Ensure proper configuration for any MachineSets which are affected.
```
dofinn, line 81: We could prob add a helper command here?

dofinn:
Great job @elmiko :)

- do we have other alerts that may be triggered at higher severity to account for the 6 waiting time for this to trigger? Example something like multiple nodes failed to register within X minutes? Looking at the causes I can see where we have coverage in cloud provider creds, but the potential network failures may pose a more difficult diagnosis.

docs/user/alerts.md
elmiko: sure that make sense to me, what did you have in mind? (eg something for the events or logs or ?)

elmiko:
>     * do we have other alerts that may be triggered at higher severity to account for the 6 waiting time for this to trigger? Example something like multiple nodes failed to register within X minutes? Looking at the causes I can see where we have coverage in cloud provider creds, but the potential network failures may pose a more difficult diagnosis.

we don't have a higher level alert on this specific condition, but i have a feeling that a user would see the [MachineWithNoRunningPhase alert](https://github.com/openshift/machine-api-operator/blob/master/docs/user/Alerts.md#machinewithnorunningphase) before this was raised. that is also a "warning" level, but i can't see either of these being escalated to "critical" level.

elmiko:
i am reviewing the alerts for the related enhancement, and i think we decided to close this out because the user will see other alerts before this would fire, and this would not reach a critical alert level.

@dofinn does that sound accurate to you? (i don't remember why i closed it before, but i want to make sure we aren't missing any details)

dofinn:
I think we posed the question if other alerts would trigger prior to this condition. Thats all I can recall.

PR #196, opened 2021-03-22
install/90_alertrules: Drop ClusterAutoscalerOperatorDown

wking:
The cluster-version operator is responsible for complaining if the autoscaler operator's deployment is sad, so no need for the operator to handle this directly (we end up doubling up if there's an issue).  This drops the alert which was originally added in e1d802526a (#119).

CC @elmiko, because this is very similar to openshift/machine-api-operator#826.


elmiko (approved):
thanks Trevor!
/lgtm

wking:
Rebased onto master with 19e9852 -> 7439efc

wking:
@elmiko , anyone I should bump for an `approved` label?

PR #197, opened 2021-04-06
Bug 1938467: The default cluster-autoscaler should get default cpu and memory requests if user omits them

Danil-Grigorev:
We are missing requests on autoscaler pods which are deployed after `ClusterAutoscaler` resource is created in the cluster.

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
@@ -11,6 +11,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


JoelSpeed (commented):
Where did these numbers come from? Have you deployed the autoscaler to any reasonably sized cluster to see what kind of resources it actually uses? Perhaps we can skip this by asking OSD folk if they have some ideas?

elmiko:
> Where did these numbers come from? Have you deployed the autoscaler to any reasonably sized cluster to see what kind of resources it actually uses? Perhaps we can skip this by asking OSD folk if they have some ideas?

i poked around the upstream repo a bit and sadly i didn't see anything useful in the [faq](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md) or the [helm chart](https://github.com/kubernetes/autoscaler/tree/master/charts/cluster-autoscaler). asking the osd folks sounds reasonable, although i have a feeling we might need to look at a running binary and make a best judgement call.

Danil-Grigorev:
> Where did these numbers come from? Have you deployed the autoscaler to any reasonably sized cluster to see what kind of resources it actually uses? Perhaps we can skip this by asking OSD folk if they have some ideas?

@JoelSpeed We don't set the limits, only reasonable requests. The numbers I set are common across all our binaries, and tht is what I observed just after pod creation on my cluster in the dashboard.

JoelSpeed:
If these are common to other operators we are deploying then that should be relatively safe, though I do think it would be better to observe actual running pods to see what kind of footprint it has. While I realise the requests are only for scheduling, the more accurate our requests are the less likely nodes will become overcommitted.

I'm happy to go forward with these for now but we should review this at a later date

/approve

PR #200, opened 2021-04-14
Bug 1948719: update dependencies to 1.21

elmiko:
This change updates the go.mod and related vendor files to the
kubernetes 1.21 baseline. It also updates one import in the
`pkg/operator/operator.go` to allow vendoring. This was achieved by
 updating some of the entries in go.mod and then running `make vendor`.

The updates to go.mod were as follows:
```
github.com/openshift/api master
github.com/openshift/client-go master
github.com/openshift/library-go master
sigs.k8s.io/controller-runtime 0.9.0-alpha.1
```

Then ran `make vendor` to produce the update. During this process the
error in `operator.go` was discovered and fixed.

Also adds necessary changes to for controller-runtime updates.


JoelSpeed:
@elmiko Seems there's a syntax error:
```
parse 0000_50_cluster-autoscaler-operator_03_rbac.yaml: error parsing: error converting YAML to JSON: yaml: line 7: did not find expected key",
```

elmiko:
not sure why it failed there, but i found a syntax error in a different part of the file.

where did you see that failure @JoelSpeed ?

JoelSpeed:
It was in the log of the upgrade test result

pkg/operator/webhookconfig.go
```diff
@@ -72,7 +72,7 @@ func (w *WebhookConfigUpdater) Start(stopCh <-chan struct{}) error {
 	klog.Infof("Webhook configuration status: %s", op)
 
 	// Block until the stop channel is closed.
-	<-stopCh
+	stop.Done()
```
Danil-Grigorev, line 75: This is not equivalent to previous code sample, should be 
```go
<-stop.Done()
```

pkg/operator/status.go
```diff
@@ -309,10 +309,10 @@ func (r *StatusReporter) Start(stopCh <-chan struct{}) error {
 		return available, nil
 	}
 
-	err := wait.PollImmediateUntil(interval, pollFunc, stopCh)
+	err := wait.PollImmediateUntil(interval, pollFunc, stop.Done())
 
 	// Block until the stop channel is closed.
-	<-stopCh
+	stop.Done()
```
Danil-Grigorev, line 315: This is not equivalent to previous code sample, should be 
```go
<-stop.Done()
```

Danil-Grigorev (commented):
One minor change and it is lgtm

pkg/operator/webhookconfig.go
elmiko: thanks Danil, i'll fix it up

elmiko:
updated to include Danil's suggestion

PR #201, opened 2021-04-16
Bug 1938467: Fix cluster autoscaler pod requests

JoelSpeed:
These seem to be the wrong way around and I think this is now causing CI failures


alexander-demichev (approved):
/lgtm

PR #202, opened 2021-04-21
Bug 1947785: Move all webhooks to v1

alexander-demichev:
Move all webhooks to v1 because v1beta1 will be removed in 4.9.

### DIFF
--- a/pkg/operator/webhookconfig.go
+++ b/pkg/operator/webhookconfig.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"fmt"
 
-	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/pointer"
@@ -43,7 +43,7 @@ func NewWebhookConfigUpdater(mgr manager.Manager, namespace string) (*WebhookCon
 // Start creates or updates the webhook configurations then waits for the stop
 // channel to be closed.
 func (w *WebhookConfigUpdater) Start(stop context.Context) error {
-	vc := &admissionregistrationv1beta1.ValidatingWebhookConfiguration{
+	vc := &admissionregistrationv1.ValidatingWebhookConfiguration{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "admissionregistration.k8s.io/v1",
 			Kind:       "ValidatingWebhookConfiguration",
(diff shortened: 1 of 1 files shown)


JoelSpeed:
/hold This is currently broken

```
Failed to start operator: [ValidatingWebhookConfiguration.admissionregistration.k8s.io "autoscaling.openshift.io" is invalid: [webhooks[0].admissionReviewVersions: Required value: must specify one of v1, v1beta1, webhooks[1].admissionReviewVersions: Required value: must specify one of v1, v1beta1], failed waiting for all runnables to end within grace period of 30s: context deadline exceeded]
```
@alexander-demichev please fix and hold cancel when ready

pkg/operator/webhookconfig.go
```diff
@@ -78,57 +78,59 @@ func (w *WebhookConfigUpdater) Start(stopCh <-chan struct{}) error {
...
+	sideEffects := admissionregistrationv1.SideEffectClassNone
 
-	webhooks := []admissionregistrationv1beta1.ValidatingWebhook{
+	webhooks := []admissionregistrationv1.ValidatingWebhook{
 		{
-			Name: "clusterautoscalers.autoscaling.openshift.io",
-			ClientConfig: admissionregistrationv1beta1.WebhookClientConfig{
-				Service: &admissionregistrationv1beta1.ServiceReference{
+			Name:                    "clusterautoscalers.autoscaling.openshift.io",
+			AdmissionReviewVersions: []string{"v1", "v1beta1"},
```
JoelSpeed, line 88: IIUC we want to just support v1 going forward as v1beta1 will be removed in 1.22
```suggestion
			AdmissionReviewVersions: []string{"v1"},
```

pkg/operator/webhookconfig.go
```diff
@@ -78,57 +78,59 @@ func (w *WebhookConfigUpdater) Start(stopCh <-chan struct{}) error {
...
 					},
 				},
 			},
 		},
 		{
-			Name: "machineautoscalers.autoscaling.openshift.io",
-			ClientConfig: admissionregistrationv1beta1.WebhookClientConfig{
-				Service: &admissionregistrationv1beta1.ServiceReference{
+			Name:                    "machineautoscalers.autoscaling.openshift.io",
+			AdmissionReviewVersions: []string{"v1", "v1beta1"},
```
JoelSpeed, line 114: IIUC we want to just support v1 going forward as v1beta1 will be removed in 1.22
```suggestion
			AdmissionReviewVersions: []string{"v1"},
```

pkg/operator/webhookconfig.go
alexander-demichev: From field description: `API server will try to use the first version in the list which it supports. `. I left v1beta as a safety check, v1 will be used anyway.

JoelSpeed: I'm just wondering if there is any validation that checks these versions against a list of supported versions, eg it works in 1.21 because v1 and v1beta1 are both supported, but breaks in 1.22 because it no longer knows what v1beta1 is 🤔 

Given every cluster that we deploy this to will be 1.21 or higher (as we aren't backporting this) we can guarantee that the clusters will support `v1`,  I don't think it's necessary to keep the fallback

elmiko:
i think we might have overlapped on this, https://github.com/openshift/machine-api-operator/pull/855

should i close out mine?

PR #210, opened 2021-06-22
Bug 1973567: add csidrivers to the cluster-autoscaler cluster role

elmiko:
The autoscaler will need to watch, list, and get these for normal
operation.

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1973567


alexander-demichev (approved):
/lgtm

PR #211, opened 2021-06-23
[OCPCLOUD-1174] Add related objects to ClusterOperator

elmiko:
This change adds the autoscaler specific resources to the
ClusterOperator manifest.


alexander-demichev (approved):
/lgtm

PR #213, opened 2021-07-07
[OCPCLOUD-923] add alerts for memory and cpu core limits

elmiko:
This change adds the alerts for when the cluster autoscaler is unable to
scale out due to reaching cpu or memory limits. It also updates the
alert documents.

ref: https://issues.redhat.com/browse/OCPCLOUD-923

### DIFF
--- a/docs/user/alerts.md
+++ b/docs/user/alerts.md
@@ -59,3 +59,55 @@ should investigate the logs associated with your cloud provider controllers and
 the Machine API resources to discover the root cause. For more information on
 why nodes, or machines, might not become ready please see the
 [Machine API FAQ](https://github.com/openshift/machine-api-operator/blob/master/FAQ.md).
+
+## ClusterAutoscalerUnableToScaleCPULimitReached
+The number of total cores in the cluster has exceeded the maximum number set on the
+cluster autoscaler. This is calculated by summing the cpu capacity for all nodes
+in the cluster and comparing that number against the maximum cores value set for the
+cluster autoscaler (default 320000 cores).
+
+### Query
+```
+# for: 15m
+cluster_autoscaler_cluster_cpu_current_cores >= cluster_autoscaler_cpu_limits_cores{direction="maximum"}
+```
+
+### Possible Causes
+* Too many nodes have been created in the cluster.
+* Nodes of larger than expected size have joined the cluster.
+* Maximum CPU limit on the ClusterAutoscaler is set too low.
+
+### Resolution
+This alert is indicating that the cluster autoscaler is unable to continue scaling out. Depending
+on your needs and resources this alert may indicate action is required. If you require more
+resources in your cluster, a simple solution is to increase the maximum core count in your
+ClusterAutoscaler. If you do not need more resources in your cluster, this condition is
+non-harmful to the cluster and the autoscaler will continue to function as normal, with the
+exception of creating new nodes. The cluster autoscaler will resume its scale out functionality
+once the number of cores in the cluster is fewer than the maximum.
+
+## ClusterAutoscalerUnableToScaleMemoryLimitReached
+The number of total bytes of RAM in the cluster has exceeded the maximum number set on
+the cluster autoscaler. This is calculated by summing the memory capacity for all nodes
+in the cluster and comparing that number against the maximum memory bytes value set
+for the cluster autoscaler (default 6400000 gigabytes).
+
+### Query
+```
+# for: 15m
+cluster_autoscaler_cluster_memory_current_bytes >= cluster_autoscaler_memory_limits_bytes{direction="maximum"}
+```
+
+### Possible Causes
+* Too many nodes have been created in the cluster.
+* Nodes of larger than expected size have joined the cluster.
+* Maximum memory limit on the ClusterAutoscaler is set too low.
+
+### Resolution
+This alert is indicating that the cluster autoscaler is unable to continue scaling out. Depending
+on your needs and resources this alert may indicate action is required. If you require more
+resources in your cluster, a simple solution is to increase the maximum memory bytes in your
+ClusterAutoscaler. If you do not need more resources in your cluster, this condition is
+non-harmful to the cluster and the autoscaler will continue to function as normal, with the
+exception of creating new nodes. The cluster autoscaler will resume its scale out functionality
+once the amount of bytes of RAM in the cluster is fewer than the maximum.
--- a/pkg/controller/clusterautoscaler/monitoring.go
+++ b/pkg/controller/clusterautoscaler/monitoring.go
@@ -191,6 +191,29 @@ func (r *Reconciler) AutoscalerPrometheusRule(ca *autoscalingv1.ClusterAutoscale
 								"message": "Cluster Autoscaler is reporting that the cluster is not ready for scaling",
 							},
 						},
+						{
+							Alert: "ClusterAutoscalerUnableToScaleCPULimitReached",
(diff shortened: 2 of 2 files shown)


elmiko:
this will need to wait until we do the 1.22 rebase of the cluster autoscaler to pick up the new metrics sources.
/hold

docs/user/alerts.md
```diff
@@ -59,3 +59,55 @@ should investigate the logs associated with your cloud provider controllers and
...
+* Too many nodes have been created in the cluster.
+* Nodes of larger than expected size have joined the cluster.
+* Maximum CPU limit on the ClusterAutoscaler is set too low.
+
+### Resolution
+This alert is indicating that the cluster autoscaler is unable to continue scaling out. Depending
+on your needs and resources this alert may indicate action is required. If you require more
+resources in your cluster, a simple solution is to increase the maximum core count in your
+ClusterAutoscaler. If you do not need more resources in your cluster, this condition is
+non-harmful to the cluster and the autoscaler will continue to function as normal, with the
```
RiRa12621, line 85: Should a potentially non-harmful condition maybe drop down to info instead of warning?

elmiko: i was curious about the same thing myself. if the autoscaler has reached its maximum (cpu or memory), it's definitely not an emergency in the cluster, but it might be something that a cluster admin would want to know since it could cause unexpected delays for users (pending pods and whatnot). does that sound more like an info level?

RiRa12621: I'll take a peak what we have around pending pods already. If that is covered with an alert itself then an info level here should be good enough as supporting information. It still pops up in the console

elmiko: awesome, thanks!

dofinn:
awesome job on this @elmiko +1

docs/user/alerts.md
RiRa12621: yeah i think 
```
   - alert: KubePodNotReady
      annotations:
        description: Pod {{ $labels.namespace }}/{{ $labels.pod }} has been in a non-ready state for longer than 15 minutes.
        summary: Pod has been in a non-ready state for more than 15 minutes.
      expr: |
        sum by (namespace, pod) (
          max by(namespace, pod) (
            kube_pod_status_phase{namespace=~"(openshift-.*|kube-.*|default|logging)",job="kube-state-metrics", phase=~"Pending|Unknown"}
          ) * on(namespace, pod) group_left(owner_kind) topk by(namespace, pod) (
            1, max by(namespace, pod, owner_kind) (kube_pod_owner{owner_kind!="Job"})
          )
        ) > 0
      for: 15m
      labels:
        severity: warning
        ```
        should be covering that case good enough

Comment (elmiko) [reply docs/user/alerts.md:85]: thanks for the confirmation @RiRa12621 , i will drop these to info level

**Comment by elmiko** [comment]
updated to drop severity from warning to info

**Comment by RiRa12621** [comment]
Lgtm from our end

**Review by michaelgugino — APPROVED** [review]
/lgtm

Info is definitely the right severity level for this one.

File: docs/user/alerts.md [hunk 2]
```diff
@@ -59,3 +59,55 @@ should investigate the logs associated with your cloud provider controllers and
 the Machine API resources to discover the root cause. For more information on
 why nodes, or machines, might not become ready please see the
 [Machine API FAQ](https://github.com/openshift/machine-api-operator/blob/master/FAQ.md).
+
+## ClusterAutoscalerUnableToScaleCPULimitReached
+The number of total cores in the cluster has exceeded the maximum number set on the
+cluster autoscaler. This is calculated by summing the cpu capacity for all nodes
+in the cluster and comparing that number against the maximum cores value set for the
+cluster autoscaler (default 320000 cores).
```
JoelSpeed, line 67: Is this default correct? Seems absurdly high? Do we have a source for that value? If so, is it worth adding a link inline?

docs/user/alerts.md
```diff
@@ -59,3 +59,55 @@ should investigate the logs associated with your cloud provider controllers and
...
+ClusterAutoscaler. If you do not need more resources in your cluster, this condition is
+non-harmful to the cluster and the autoscaler will continue to function as normal, with the
+exception of creating new nodes. The cluster autoscaler will resume its scale out functionality
+once the number of cores in the cluster is fewer than the maximum.
+
+## ClusterAutoscalerUnableToScaleMemoryLimitReached
+The number of total bytes of RAM in the cluster has exceeded the maximum number set on
+the cluster autoscaler. This is calculated by summing the memory capacity for all nodes
+in the cluster and comparing that number against the maximum memory bytes value set
+for the cluster autoscaler (default 6400000 gigabytes).
```
JoelSpeed, line 93: Same as above, where did this figure come from?

JoelSpeed (commented):
This seems good to me, just curious about the default values mentioned, PTAL

docs/user/alerts.md
elmiko: this is just the default if you set nothing on the cluster autoscaler, see [this faq entry](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca), also we do allow [not setting those limits in a ClusterAutoscaler](https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/apis/autoscaling/v1/clusterautoscaler_types.go#L77-L94)

docs/user/alerts.md
elmiko: same as above

elmiko:
reminder: update https://github.com/openshift/enhancements/pull/538 when this merges

docs/user/alerts.md
wanghaoran1988: Could you add a description how you can get the current maxmum settings? like "You can run this to get what's the maximum cores setup in this cluster".

JoelSpeed: @wanghaoran1988 Does this need to happen before we merge or could that follow in a later release? I'm happy with this personally and would prefer to merge it today so its in before feature freeze, but Mike is out on PTO today so wouldn't have time to update this before FF

wanghaoran1988: I'm ok with a follow release.

elmiko:
this PR technically needs the metric to be present, we will pick it up during the 1.22 rebase. i am removing the hold as we don't believe this will break anything by merging early.
/hold cancel

elmiko:
seeing this error several times in the logs
```
 03   11966 framework.go:128] error querying api for ValidatingWebhookConfiguration: no matches for kind "ValidatingWebhookConfiguration" in version "admissionregistration.k8s.io/v1beta1", retrying...
E0725 14:49:30.977789   11966 framework.go:128] error querying api for ValidatingWebhookConfiguration: no matches for kind "ValidatingWebhookConfiguration" in version "admissionregistration.k8s.io/v1beta1", retrying... 
```

doesn't seem related to this PR, but makes me wonder if that is one of the types that got upgraded to `v1` version?

elmiko:
seems like yes, https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22

JoelSpeed:
The webhook test is a known perma failure at the moment, we have a PR to our tests updating the webhook versions but other tests are flaking causing it not have merged yet

Having reviewed the failures, I don't think any of them are related to this PR, this should be safe to merge 

elmiko:
thanks Joel, i was just looking at that webhook version failure lol

JoelSpeed:

Re overide based on the above comment

PR #215, opened 2021-07-26
Bug 1986090: Do not recreate CA deployment when CA CR is being deleted

JoelSpeed:
This is breaking our tests which use foreground deletion. The CAO fights the GC to delete and recreate the deployment, ensuring that the CA CR never goes away.

I have manually verified that this fixes the broken cluster autoscaler tests we are currently seeing in CI.

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
@@ -180,13 +180,25 @@ func (r *Reconciler) Reconcile(_ context.Context, request reconcile.Request) (re
 		return reconcile.Result{}, err
 	}
 
-	_, err = r.GetAutoscaler(ca)
+	existingDeployment, err := r.GetAutoscaler(ca)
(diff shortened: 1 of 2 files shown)


elmiko (commented):
this looks like a good solution, just curious if we could add a unit test to check the deletion time stamp part of this?

JoelSpeed:
> just curious if we could add a unit test to check the deletion time stamp part of this?

Reasonable request! I did have a look and our unit tests right now are a bit not great in this area, so it won't be trivial to add a test for this. I will look into adding a unit test for this tomorrow though.

elmiko:
> Reasonable request! I did have a look and our unit tests right now are a bit not great in this area, so it won't be trivial to add a test for this. I will look into adding a unit test for this tomorrow though.

sounds good, i was hoping we could slip this into the reconcile test but maybe that was optimistic of me ;)

alexander-demichev (approved):
/approve

JoelSpeed:

The tests are broken, I've manually verified this, once this merged we can get the rest of the tests fixed

PR #216, opened 2021-07-29
Bug 1988032: add cvo ha annotation to tombstones

elmiko:
This change adds the annotation
`include.release.openshift.io/self-managed-high-availability: "true"`
to the tombstones. This is needed to ensure that the CVO will process
them.

### DIFF
--- a/install/99_tombstones.yaml
+++ b/install/99_tombstones.yaml
@@ -4,12 +4,14 @@ metadata:
   name: cluster-autoscaler-operator-ca
   namespace: openshift-machine-api
   annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
     release.openshift.io/delete: "true"
 ---
-apiVersion: monitoring.coreos.com
+apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
   name: cluster-autoscaler-operator-rules
   namespace: openshift-machine-api
   annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
     release.openshift.io/delete: "true"


jottofar:
e2e-aws is getting:

E0729 18:21:05.029763       1 task.go:112] error running apply for prometheusrule "openshift-machine-api/cluster-autoscaler-operator-rules" (286 of 719): failed to get resource type: no matches for kind "PrometheusRule" in version "monitoring.coreos.com"

No version on apiVersion is perhaps the issue:

apiVersion: monitoring.coreos.com

elmiko:
as i understand it, this type was removed in 4.7, is there some other way i should be specifying this?

re: https://issues.redhat.com/browse/OCPCLOUD-1153 for reference on the apiVersion

elmiko:
hmm, so looking back through the historical versions of the MAO repo, i am not seeing the original file which created this object. perhaps we have an error in trying to delete this?

elmiko:
ok, i think i see the error. the version should be `monitoring.coreos.com/v1`, according to this file where it was originally created https://github.com/openshift/cluster-autoscaler-operator/blob/release-4.3/install/09_alertrules.yaml

jottofar:
Maybe I misunderstand your comment but I think your manifests should specify:
`apiVersion: monitoring.coreos.com/v1`

That is I think it's complaining because you are not specifying `/v1`.
I'm running a 4.8 cluster and CVO is sync'ing other manifests with that apiVersion just fine.

The manifest from release 4.7.17 that I based the bug on is 0000_50_cluster-autoscaler-operator_09_alertrules.yaml

elmiko:
thanks for the confirmation @jottofar , updated to correct the `apiVersion`

Fedosin (approved):
/lgtm

PR #224, opened 2021-11-02
Bug 2001027: add topology.ebs.csi.aws.com/zone to ignored labels

elmiko:
when balancing similar nodes, the autoscaler will look for differences
in labels as one of its checks for similarity. by default it will ignore
the well known kubernetes zone label (topology.kubernetes.io/zone), but
the aws-ebs-csi-driver will add a unique label for its zone,
`topology.ebs.csi.aws.com/zone`. this label will eventually be
deprecated in favor of the common well known topology label. for more
information about this see the references.

ref: kubernetes/autoscaler#3230
ref: kubernetes-sigs/aws-ebs-csi-driver#729

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler.go
@@ -52,6 +52,7 @@ const (
 	BalanceSimilarNodeGroupsArg     AutoscalerArg = "--balance-similar-node-groups"
 	IgnoreDaemonsetsUtilization     AutoscalerArg = "--ignore-daemonsets-utilization"
 	SkipNodesWithLocalStorage       AutoscalerArg = "--skip-nodes-with-local-storage"
+	BalancingIgnoreLabel            AutoscalerArg = "--balancing-ignore-label"
 )
 
 // AutoscalerArgs returns a slice of strings representing command line arguments
@@ -92,6 +93,12 @@ func AutoscalerArgs(ca *v1.ClusterAutoscaler, cfg *Config) []string {
 
 	if ca.Spec.BalanceSimilarNodeGroups != nil {
 		args = append(args, BalanceSimilarNodeGroupsArg.Value(*ca.Spec.BalanceSimilarNodeGroups))
+		// if we are enabling the balance similar node groups option, we need to ignore
+		// openshift specific labels that prevent the default balancing behavior.
+		// TODO (elmiko) remove this label once it has been fully deprecated from the aws-ebs-csi-driver
+		if *ca.Spec.BalanceSimilarNodeGroups == true {
+			args = append(args, BalancingIgnoreLabel.Value("topology.ebs.csi.aws.com/zone"))
+		}
 	}
 
 	if ca.Spec.IgnoreDaemonsetsUtilization != nil {


enxebre:
Can you please link to where is OCP propagating this to nodes?

elmiko:
@enxebre we add it in the machine actuator, for example:

aws - https://github.com/openshift/cluster-api-provider-aws/blob/master/pkg/actuators/machine/reconciler.go#L411
vsphere - https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/vsphere/reconciler.go#L401

elmiko:
oh sorry, you said "nodes". i will have to dig a little more for that. i was looking at the machine objects in my testing, i will look at the nodes too.

elmiko:
just double checked the node objects and i do not see this propagated on the node. it seems like it is just on the machine object.

enxebre:
>just double checked the node objects and i do not see this propagated on the node. it seems like it is just on the machine object.

So `FindSimilarNodeGroups` use the labels in the nodes right? so how can this label change anything?

elmiko:
yeah, that's a good point. i had a hunch that these labels were getting propagated to the simulator, but i think i need to dig through the autoscaler code a little more. i think this would only happen during a scale from zero situation where the autoscaler is using the template info to simulate what the node would look like.

elmiko:
i am noticing that our nodes carry this label `failure-domain.beta.kubernetes.io/zone`, which is deprecated as of 1.17 in favor of the `topology.kubernetes.io/zone` label. but, it does not look like the autoscaler is taking that label into account. i am going to try reworking this pr with that label, if it works then perhaps we should propose this in the upstream.

enxebre:
>i am noticing that our nodes carry this label failure-domain.beta.kubernetes.io/zone, which is deprecated as of 1.17 in favor of the topology.kubernetes.io/zone label. but, it does not look like the autoscaler is taking that label into account. i am going to try reworking this pr with that label, if it works then perhaps we should propose this in the upstream.

That was fixed a while ago https://github.com/kubernetes/autoscaler/pull/2848, I wouldn't expect it to be related with the linked bug.

elmiko:
> That was fixed a while ago [kubernetes/autoscaler#2848](https://github.com/kubernetes/autoscaler/pull/2848), I wouldn't expect it to be related with the linked bug.

ok, cool. i missed that in my search. i'm not seeing how these nodes are different from kube api perspective, but i have observed the behavior on a live cluster. i know we are failing to balance properly when the AZ is different between 2 machinesets. i'll keep digging.

enxebre:
fwiw If you are able to reproduce it consistently running autoscaler remotely with the debugger might be insightful

elmiko:
yeah, that's my next step. was trying to see if i could figure out from the code, but it seems like i've run into a wall here. lol

elmiko:
this may be an aws specific thing, although something similar could happen in other clouds. it appears that it is failing on this label `topology.ebs.csi.aws.com/zone`. i'm doing a little more testing to confirm.

elmiko:
i'm going to do a little more research to see if there are a list of these labels for the csi drivers, if so we can donate a patch upstream to add these labels. otherwise, we should probably detect the platform and then add the labels to our ignore list in the CAO.

does that sound good to you @enxebre ?

elmiko:
it appears this is where the label originates from, https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/773

elmiko:
given the discussion here, https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/729 , i think there is a case to be made for adding this to the upstream autoscaler until the aws driver is able to deprecate that label. i can make a pr and bring it up with them.

enxebre:
Upstream first and back port when it's merged make sense to me, It's tracked here https://github.com/kubernetes/autoscaler/issues/3230
Otherwise will need to ensure that label is passed through the ignore-labels flag.

elmiko:
> Upstream first and back port when it's merged make sense to me, It's tracked here [kubernetes/autoscaler#3230](https://github.com/kubernetes/autoscaler/issues/3230) Otherwise will need to ensure that label is passed through the ignore-labels flag as you were intending in this PR.

that thread looks interesting, i'm going to need to read through it a little more carefully to see if upstream will want this change. it seems like they are pushing the `--balancing-ignore-label` approach. which is understandable, i know they don't want provider specific code in the core parts of the autoscaler.

> Is there any e2e addition we can do to cover this?

we do have a test for node balancing now, but i think we need to revisit it and make sure that we are placing the machinesets in different zones, if possible.

enxebre:
>we do have a test for node balancing now, but i think we need to revisit it and make sure that we are placing the machinesets in different zones, if possible.

Yeh I mean the test didn't highlighted the issue. Ideally we should be able to reproduce the scenario and get it green in CI.

elmiko:
> Yeh I mean the test didn't highlighted the issue. Ideally we should be able to reproduce the scenario and get it green in CI.

agreed, i think the complicated part for setting this up as a test is that currently this is aws only and we will need to care about the zone and currently we don't do anything about that. i'm sure we can work it out, but i think it will be a little complicated.

elmiko:
i have updated this pr to reflect the label we like to ignore, if the upstream autoscaler does not want to accept this into the builtin ignored labels we can proceed with this pr.

i think the only update i would like to make that i can't quite figure out is how to add selective logic to only add this label for aws clusters.

elmiko:
i've talked with upstream and the preferred solution here is to use a custom nodegroupset processor. i am going to propose a patch to the upstream that will add a new processor for cluster-api. that will allow us to properly handle these cases from the autoscaler in the future.

PR #226, opened 2021-11-19
OCPCLOUD-1278 - add scale-down-utilization-threshold to CAO

SamuelStuchly:
Add a new field in the CAO `ClusterAutoscaler` resource in the scale down section.
Plumb through the logic from the CRD to the autoscaler deployment command line flag.


pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -128,4 +128,8 @@ type ScaleDownConfig struct {
 	// How long a node should be unneeded before it is eligible for scale down
 	// +kubebuilder:validation:Pattern=([0-9]*(\.[0-9]*)?[a-z]+)+
 	UnneededTime *string `json:"unneededTime,omitempty"`
+
+	// Node utilization level, defined as sum of requested resources divided by capacity, below which a node can be considered for scale down
+	// +kubebuilder:validation:Pattern=(0.[0-9]+)
```
JoelSpeed, line 133: Is there any validation on this flag done by the cluster autoscaler? Can the threshold ever be set to `0` or `1`? As neither of these are currently allowed by this pattern

SamuelStuchly: Yes the validation is done in pkg/controller/clusterautoscaler/validator.go. (Cant seem to link the changes.) 

Since this threshold represents threshold below which scale down can be initiated. Setting threshold to 0 or 1 does not make much sense.

elmiko: i checked the autoscaler itself, and it does not have a validation for this that i could find, but i did see one unit test that was using a value above `1`. which makes me wonder if there is an over-provisioning case we need to account for?

basically, could a user set it to `1.5` with the expectations that the node could be over committed on the requested resources. i'm not sure if kube would allow this.

but, all that said, i'm not even sure the case for a value greater than `1` is valid. i'm fine with this regex the way it is.

pkg/controller/clusterautoscaler/validator.go
```diff
@@ -136,6 +137,13 @@ func (v *Validator) validateScaleDownConfig(sd *autoscalingv1.ScaleDownConfig) u
 			errs = append(errs, fmt.Errorf("ScaleDown.%s: %v", name, err))
 		}
 	}
+	utilizationThreshold, err := strconv.ParseFloat(*sd.UtilizationThreshold, 64)
+	if err != nil {
+		errs = append(errs, errors.New("ScaleDown.UtilizationThreshold must be a string representing float value."))
+	}
+	if utilizationThreshold <= float64(0) || utilizationThreshold >= float64(1) {
+		errs = append(errs, errors.New("ScaleDown.UtilizationThreshold must be a value between 0 and 1."))
+	}
```
elmiko, line 146: i think we should add a unit test to exercise this validation

elmiko (commented):
looks mostly good, i think we should add one more unit test for the validator then i would be happy to merge.

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
```diff
@@ -36,18 +36,19 @@ const (
...
-	MaxPodGracePeriod      int32 = 60
-	MaxNodesTotal          int32 = 100
-	CoresMin               int32 = 16
-	CoresMax               int32 = 32
-	MemoryMin              int32 = 32
-	MemoryMax              int32 = 64
-	NvidiaGPUMin           int32 = 4
-	NvidiaGPUMax           int32 = 8
+	ScaleDownUnneededTime               = "10s"
+	ScaleDownUtilizationThreshold       = "0.4"
```
JoelSpeed, line 40: What is the default value for this in the autoscaler itself? I was under the impression it was 0.5

elmiko: that's correct

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
elmiko: @SamuelStuchly here is where i saw the utilization threshold set over 1,

https://github.com/openshift/kubernetes-autoscaler/blob/master/cluster-autoscaler/cloudprovider/ovhcloud/ovh_cloud_node_group_test.go#L120

pkg/controller/clusterautoscaler/validator.go
elmiko: @SamuelStuchly would you mind adding a unit test for this?

elmiko (commented):
i'd just like to see a unit test for the new validation, then i'm good to merge.

pkg/controller/clusterautoscaler/validator.go
SamuelStuchly: Will do.

elmiko (commented):
thanks Sam!
/lgtm

elmiko (commented):
/lgtm

PR #228, opened 2021-11-20
Bug 2025230: Change ClusterAutoscalerUnschedulablePods severity to info

cblecker:
This alert is informational for the cluster operator, and doesn't describe an actual problematic condition that requires human action.

Per the documentation:
> In many cases this alert is normal and expected depending on the configuration of the autoscaler.

As such, lowering the severity of this to "info".


elmiko (commented):
thanks @cblecker 
/approve

PR #229, opened 2021-11-20
Migrate alert documentation to runbooks

cblecker:
This PR migrates the documentation for these alerts to the github.com/openshift/runbooks repo.

This shouldn't merge until after the following PRs:
https://github.com/openshift/runbooks/pull/26
https://github.com/openshift/cluster-autoscaler-operator/pull/228

/hold


elmiko (commented):
would it be ok to leave the alerts doc in this repo as well?

our team uses this document as our first location to add more information, i'm totally ok with adding the link to the run books, but i think it's important to leave the doc here as well.

cblecker:
@elmiko My concern would be things getting out of sync. The runbooks should be the canonical source for this type of documentation. I'm open if you have any suggestions to try and keep things in sync.

elmiko:
i'd like to talk with our team about this a little bit since we maintain a document like this in our MAO repo as well. if the guidance here is put everything in the runbooks then i definitely agree we don't want to get out of sync. maybe we need to have all our metrics docs point to the runbooks.

cblecker:
Sounds great! Let me know what your team thinks about this :)

PR #234, opened 2021-12-10
Update makefile to support podman

SamuelStuchly:
Update make targets to be able to run in container with podman and docker, as well as run without container.


alexander-demichev (approved):
/approve

PR #235, opened 2021-12-16
Library go bump

eggfoobar:
Bumping the library-go dependency to latest, this should take advantage of leader election changes for SNO clusters proposed in this [library-go PR](https://github.com/openshift/library-go/pull/1242) and performance improvements in this [PR](https://github.com/openshift/library-go/pull/1235).

Changes:
  - updated library-go to latest
  - added logic to fetch default leader election configs from library-go

PR #236, opened 2021-12-21
Bug 2034484: fix: missed rbac and latest library-go

eggfoobar:
updated to include missed rbac permission for infrastructure resource
updated library-go to latest for fix in pr https://github.com/openshift/library-go/pull/1273

Signed-off-by: ehila <ehila@redhat.com>

### DIFF
--- a/install/03_rbac.yaml
+++ b/install/03_rbac.yaml
@@ -110,6 +110,14 @@ rules:
     - update
     - patch
     - delete
+- apiGroups:
+    - config.openshift.io
+  resources:
+    - infrastructures
+  verbs:
+    - get
+    - list
+    - watch
 
 ---
 kind: RoleBinding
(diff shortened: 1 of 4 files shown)


eggfoobar:
/retitle Bug 2034484: fix: missed rbac and latest library-go
this PR pulls in the latest library-go to utilize needed performance improvements, and fixes rbac bug

install/03_rbac.yaml
```diff
@@ -110,6 +110,14 @@ rules:
     - update
     - patch
     - delete
+- apiGroups:
+    - config.openshift.io
+  resources:
+    - infrastructures
```
elmiko, line 116: i looked through the code but i couldn't see where this operator is attempting to use an infrastructure object, do we need this permission?

elmiko (commented):
this generally looks good to me, just having a little trouble linking the permission to code.

install/03_rbac.yaml
eggfoobar: Ah sorry, that was in the previous PR, it's being called here https://github.com/openshift/cluster-autoscaler-operator/pull/235/files#diff-c2a06a8d073270721be5c29036cf48fc7b1cffe05223c2f277fbe46bc967180aR46

elmiko: ahh, thank you very much. i missed that call.

Fedosin (approved):
/lgtm

PR #244, opened 2022-03-30
Bug 2070277: add leader election flags to autoscaler deployment

elmiko:
This change is being added to bring the cluster autoscaler deployment in
line with the OpenShift conventions[0] for leader election. It also adds
test for the default flags to help prevent regressions.

[0] https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md


elmiko:
seems like we'll need the blocker bug to move into VERIFIED state before we can get the bug label here

PR #252, opened 2022-09-27
OCPBUGS-1411: Bump k8s deps to 1.25

lobziik:
Also upd controller-runtime and openshift/api


elmiko (commented):
this update looks fine to me, it is possible that some of the failures we are seeing on e2e-aws-operator are related to the issues i brought up about nodes with NoSchedule taints affecting other tests.
/approve
/lgtm

elmiko:
there are some issues happening with the e2e-aws-operator test, i will run it again but looking at the output here i think we are safe to merge this. if the next test run does not complete, i will override and merge so that we can then update the vendor directory of the test package.

/hold cancel
/test e2e-aws-operator

elmiko:
this looks to be failing in a known problem area, i am confident it has passed the normal tests. i am going to override.

PR #254, opened 2022-10-05
Bug 1997396: fix an error with autoscaler alert rules

elmiko:
the new rules for the resource limits failed to include the time unit for the range vector, this change fixes that.


elmiko:
fwiw, i manually ran the entire cluster-api-actuator-pkg against this PR and it worked

elmiko:
we discussed this as a team and feel that it is important enough that we will override the failing tests. i have tested this manually and it is necessary to unblock our failing e2e-aws-operator tests.

/override ci/prow/e2e-aws
/override ci/prow/e2e-upgrade

PR #255, opened 2022-10-19
Bug 1943194: update SetGPUAcceleratorLabel to add labels for nodes

elmiko:
this change makes it so that the CAO will add labels to the machineset in a manner that will cause nodes to be created with the same labels. additionally, it adds the gpu acclerator label to those machinesets so that the autoscaler will properly evaluate the upcoming nodes.

### DIFF
--- a/pkg/controller/machineautoscaler/machinetarget.go
+++ b/pkg/controller/machineautoscaler/machinetarget.go
@@ -263,8 +263,13 @@ func (mt *MachineTarget) HasGPUCapacity() bool {
 func (mt *MachineTarget) HasGPUAcceleratorLabel() bool {
 	labels := mt.GetLabels()
 	// the value of this label is not used, only its presence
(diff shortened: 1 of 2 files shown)


JoelSpeed:
/lgtm

Keen to see QE verify this before we merge

elmiko:
@JoelSpeed agreed, should we hold this until QE can confirm for us?

JoelSpeed:
Yep, will run a retest but lets try and get QE to verify this, CC @sunzhaohua2 

/retest-required

elmiko:
i need to do some updates here, it will also involve changes to the API. we need to keep the hold here until that work is done.

elmiko:
i am closing this PR as we have decided to not update the API but instead will instruct the users no how to properly mark their MachineSets. we will also add some validation to ensure that the values passed in for GPU limits follow the rules for label naming.

PR #256, opened 2022-11-21
Update clusterautoscaler.yaml

peterlai-roboops:
Removing logVerbosity, as it doesn't exists. https://docs.openshift.com/container-platform/4.11/rest_api/autoscale_apis/clusterautoscaler-autoscaling-openshift-io-v1.html

### DIFF
--- a/examples/clusterautoscaler.yaml
+++ b/examples/clusterautoscaler.yaml
@@ -4,7 +4,6 @@ kind: "ClusterAutoscaler"
 metadata:
   name: "default"
 spec:
-  logVerbosity: 1
   balanceSimilarNodeGroups: true
   ignoreDaemonsetsUtilization: false
   skipNodesWithLocalStorage: true


elmiko:
/hold

this field definitely exists in the master branch, it might not have made the 4.11 docs, but it is present. see https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/apis/autoscaling/v1/clusterautoscaler_types.go#L55

lobziik:
Yeah, this does not exist in 4.11 and was added after 4.11 release did cut.

/lgtm cancel

elmiko:
@peterlai-roboops i would like to close this issue unless you think there is a reason we need it?

elmiko:
i am closing this issue as it has been over a month since the last response. if you feel this is an error, please re-open the issue and add some details for the reasoning.

/close

PR #257, opened 2022-11-28
WIP update SetGPUAcceleratorLabel to add labels for nodes

elmiko:
this change makes it so that the CAO will add labels to the machineset in a manner that will cause nodes to be created with the same labels. additionally, it adds the gpu acclerator label to those machinesets so that the autoscaler will properly evaluate the upcoming nodes.


elmiko:
i am proposing this change so that we can do some testing

elmiko:
i am closing this as we are going to go in a different direction with this solution. instead of having the CAO add this label to the machineset, and possibly confuse users, we will instruct users to add the field themselves.

PR #260, opened 2022-12-15
[OCPCLOUD-1702]: Update CAO to ignore platform related zone labels

odvarkadaniel:
This PR adds the ability for us to determine what provider platform are we using. With that knowledge, we can always append ignore zone labels for related platforms.
cc @elmiko

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler.go
@@ -2,6 +2,7 @@ package clusterautoscaler
 
 import (
 	"fmt"
+	"strings"
 
 	v1 "github.com/openshift/cluster-autoscaler-operator/pkg/apis/autoscaling/v1"
 )
@@ -109,6 +110,17 @@ func AutoscalerArgs(ca *v1.ClusterAutoscaler, cfg *Config) []string {
 
 	if ca.Spec.BalanceSimilarNodeGroups != nil {
 		args = append(args, BalanceSimilarNodeGroupsArg.Value(*ca.Spec.BalanceSimilarNodeGroups))
+
+		switch strings.ToLower(cfg.CloudProvider) {
+		case "alibabacloud":
+			args = append(args, BalancingIgnoreLabelArg.Value("topology.diskplugin.csi.alibabacloud.com/zone"))
+		case "aws":
+			args = append(args, BalancingIgnoreLabelArg.Value("topology.ebs.csi.aws.com/zone"))
+		case "azure":
+			args = append(args, BalancingIgnoreLabelArg.Value("topology.disk.csi.azure.com/zone"))
+		case "gcp":
+			args = append(args, BalancingIgnoreLabelArg.Value("topology.gke.io/zone"))
+		}
 	}
 
 	if ca.Spec.IgnoreDaemonsetsUtilization != nil {
--- a/pkg/controller/clusterautoscaler/clusterautoscaler.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler.go
@@ -69,6 +69,95 @@ const (
 	LeaderElectRetryPeriodArg        AutoscalerArg = "--leader-elect-retry-period"
 )
 
+// The following values are for cloud providers which have not yet created specific nodegroupset processors.
+// These values should be removed and replaced in the event that one of the cloud providers creates a nodegroupset processor.
+
+// AWS cloud provider ignore labels for the autoscaler.
+const (
+	// AwsIgnoredLabelEksctlInstanceId  is a label used by eksctl to identify instances.
+	AwsIgnoredLabelEksctlInstanceId = "alpha.eksctl.io/instance-id"
+
+	// AwsIgnoredLabelEksctlNodegroupName is a label used by eksctl to identify "node group" names.
+	AwsIgnoredLabelEksctlNodegroupName = "alpha.eksctl.io/nodegroup-name"
+
+	// AwsIgnoredLabelEksNodegroup is a label used by eks to identify "node group".
+	AwsIgnoredLabelEksNodegroup = "eks.amazonaws.com/nodegroup"
+
+	// AwsIgnoredLabelK8sEniconfig is a label used by the AWS CNI for custom networking.
+	AwsIgnoredLabelK8sEniconfig = "k8s.amazonaws.com/eniConfig"
+
+	// AwsIgnoredLabelLifecycle is a label used by the AWS for spot.
+	AwsIgnoredLabelLifecycle = "lifecycle"
+
+	// AwsIgnoredLabelEbsCsiZone is a label used by the AWS EBS CSI driver as a target for Persistent Volume Node Affinity.
+	AwsIgnoredLabelEbsCsiZone = "topology.ebs.csi.aws.com/zone"
+)
+
+// IBM cloud provider ignore labels for the autoscaler.
+const (
+	// IbmcloudIgnoredLabelWorkerId is a label used by the IBM Cloud Cloud Controler Manager.
+	IbmcloudIgnoredLabelWorkerId = "ibm-cloud.kubernetes.io/worker-id"
+
+	// IbmcloudIgnoredLabelVpcBlockCsi is a label used by the IBM Cloud CSI driver as a target for Persisten Volume Node Affinity.
+	IbmcloudIgnoredLabelVpcBlockCsi = "vpc-block-csi-driver-labels"
+
+	// IbmcloudIgnoredLabelVpcInstanceId on IBM Cloud when a VPC is in use.
+	IbmcloudIgnoredLabelVpcInstanceId = "ibm-cloud.kubernetes.io/vpc-instance-id"
+)
+
+// GCP cloud provider ignore labels for the autoscaler.
+const (
+	// GceIgnoredLabelGkeZone label is used to specify the zone of the instance.
+	GceIgnoredLabelGkeZone = "topology.gke.io/zone"
+)
+
+// Azure cloud provider ignore labels for the autoscaler.
+const (
+	// AzureDiskTopologyKey is the topology key of Azure Disk CSI driver.
+	AzureDiskTopologyKey = "topology.disk.csi.azure.com/zone"
+
+	// AzureNodepoolLegacyLabel is a label specifying which Azure node pool a particular node belongs to.
+	AzureNodepoolLegacyLabel = "agentpool"
+
+	// AzureNodepoolLabel is an AKS label specifying which nodepool a particular node belongs to.
+	AzureNodepoolLabel = "kubernetes.azure.com/agentpool"
+)
+
+// Alibaba cloud provider ignore labels for the autoscaler.
+const (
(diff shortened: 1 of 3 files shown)


pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -109,6 +110,17 @@ func AutoscalerArgs(ca *v1.ClusterAutoscaler, cfg *Config) []string {
...
+
+		switch strings.ToLower(cfg.CloudProvider) {
+		case "alibabacloud":
+			args = append(args, BalancingIgnoreLabelArg.Value("topology.diskplugin.csi.alibabacloud.com/zone"))
+		case "aws":
+			args = append(args, BalancingIgnoreLabelArg.Value("topology.ebs.csi.aws.com/zone"))
+		case "azure":
+			args = append(args, BalancingIgnoreLabelArg.Value("topology.disk.csi.azure.com/zone"))
+		case "gcp":
+			args = append(args, BalancingIgnoreLabelArg.Value("topology.gke.io/zone"))
```
JoelSpeed, line 122: Wonder if we should make the various cloud provider label names constants that we can group per platform? I feel like if we need to add more this could become confusing if they are embedded on individual lines

elmiko: +1, i think that's a nice upgrade

elmiko (commented):
this is a good direction, i think we need to also bring the other aws and ibm labels that i referenced in this PR https://github.com/kubernetes/autoscaler/pull/5148

odvarkadaniel:
In the latest commit, I have added the rest of the labels that @elmiko pointed in his issue.
Also grouped the ignore labels per cloud provider and moved that into a function so we can work with that better in the future.

pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -131,6 +219,8 @@ func AutoscalerArgs(ca *v1.ClusterAutoscaler, cfg *Config) []string {
 		args = append(args, VerbosityArg.Value(cfg.Verbosity))
 	}
 
+	fmt.Println(args)
```
JoelSpeed, line 222: Debug, remove before merge

pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -68,6 +69,90 @@ const (
...
+
+	// AwsIgnoredLabelEbsCsiZone is a label used by the AWS EBS CSI driver as a target for Persistent Volume Node Affinity.
+	AwsIgnoredLabelEbsCsiZone = "topology.ebs.csi.aws.com/zone"
+)
+
+// The following values are for cloud providers which have not yet created specific nodegroupset processors.
+// These values should be removed and replaced in the event that one of the cloud providers creates a nodegroupset processor.
+const (
+	// AlicloudIgnoredLabelCsiZone is a label used by the Alibaba Cloud CSI driver as a target for Persistent Volume Node Affinity.
+	AlicloudIgnoredLabelCsiZone = "topology.diskplugin.csi.alibabacloud.com/zone"
```
JoelSpeed, line 98: Do we want alibaba in their own const group?

pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -68,6 +69,90 @@ const (
...
+	IbmcloudIgnoredLabelVpcInstanceId = "ibm-cloud.kubernetes.io/vpc-instance-id"
+)
+
+// GCP cloud provider ignore labels for the autoscaler.
+const (
+	// GceIgnoredLabelGkeZone label is used to specify the zone of the instance.
+	GceIgnoredLabelGkeZone = "topology.gke.io/zone"
+)
+
+const (
```
JoelSpeed, line 116: Comment on the const group for consistency?

pkg/controller/clusterautoscaler/clusterautoscaler.go
odvarkadaniel: Alibaba is mentioned in the PR Mike mentioned in earlier comment. What solution are you thinking about?

JoelSpeed: Just all the other clouds are grouped individually, you have AWS consts, then azure, then GCP, but IBM and baba are mixed together?

odvarkadaniel: Oh yes, missed that. Sorry. Fixed.

pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -68,6 +69,93 @@ const (
...
+	AzureDiskTopologyKey = "topology.disk.csi.azure.com/zone"
+
+	// AzureNodepoolLegacyLabel is a label specifying which Azure node pool a particular node belongs to.
+	AzureNodepoolLegacyLabel = "agentpool"
+
+	// AzureNodepoolLabel is an AKS label specifying which nodepool a particular node belongs to.
+	AzureNodepoolLabel = "kubernetes.azure.com/agentpool"
+)
+
+const (
```
JoelSpeed, line 125: Comment for const group for consistency :)

pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -68,6 +69,93 @@ const (
...
+
+	// AwsIgnoredLabelLifecycle is a label used by the AWS for spot.
+	AwsIgnoredLabelLifecycle = "lifecycle"
+
+	// AwsIgnoredLabelEbsCsiZone is a label used by the AWS EBS CSI driver as a target for Persistent Volume Node Affinity.
+	AwsIgnoredLabelEbsCsiZone = "topology.ebs.csi.aws.com/zone"
+)
+
+// The following values are for cloud providers which have not yet created specific nodegroupset processors.
+// These values should be removed and replaced in the event that one of the cloud providers creates a nodegroupset processor.
```
JoelSpeed, line 95: Do we want to make this comment consistent with the others?

JoelSpeed (commented):
Couple of nits, otherwise LGTM

pkg/controller/clusterautoscaler/clusterautoscaler.go
odvarkadaniel: Would you prefer to have this exact comment placed for GCP, Azure and Alibaba?

JoelSpeed: Well, does this comment actually make sense in isolation? Maybe have it at the top, above all of the constant groups? As far as I know, none of these have specific nodegroupset processors because we are using them with the CAPI processor instead?

odvarkadaniel: Yes, that is true. I will then place it at the top of the const declarations.

pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -68,6 +69,95 @@ const (
...
+// Alibaba cloud provider ignore labels for the autoscaler.
+const (
+	// AlicloudIgnoredLabelCsiZone is a label used by the Alibaba Cloud CSI driver as a target for Persistent Volume Node Affinity.
+	AlicloudIgnoredLabelCsiZone = "topology.diskplugin.csi.alibabacloud.com/zone"
+)
+
+// AppendBasicIgnoreLabels appends ignore labels for specific cloud provider to the arguments
+// so the autoscaler can use these labels without the user having to input them manually.
+func appendBasicIgnoreLabels(args []string, cfg *Config) []string {
+	switch strings.ToLower(cfg.CloudProvider) {
```
elmiko, line 135: i think it would be better if we could use the consts from here https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go to use for checking the platform type.

elmiko (commented):
mostly looks good to me, i think we should use the constant values for the platform types though.

also, were you going to refactor this @odvarkadaniel ?

pkg/controller/clusterautoscaler/clusterautoscaler.go
odvarkadaniel: I think so as well. That is exactly what I proposed to you on Slack + some other things we could to do make our work easier in the future. I think it might be better for me to do that in separate PR on top of this change, wdyt?
I tried with this approach at first, but it turned out that there might be changes needed to some other parts of the CAO code base.

elmiko (changes requested):
i don't think this approach will work the way it is. the cloud provider that is being used to detect the platform appears to be the value that is passed to the cluster-autoscaler binary as its `--cloud-provider` flag, which should always be `clusterapi`.

we need to detect the paltform type from the [platformType](https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go#L60) in the infrastructure object, which means we will need to add the necessary rbac to read it, and then add a function to get the platform type from there.

see the following for current provider stuff:

default value
https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/operator/config.go#L42-L44

setting default
https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/operator/config.go#L143

trying to get from environment var
https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/operator/config.go#L190

odvarkadaniel:
The latest commit retrieves the `providerType` from the `Infrastructure` object. We update the `clusterAutoscaler` object within the Reconcile function. We add a field that stores the `providerType` in the Autoscaler object.

JoelSpeed:
> We update the clusterAutoscaler object within the Reconcile function. We add a field that stores the providerType in the Autoscaler object.

I don't like the idea of storing this on the `ClusterAutoscaler` CRD, can we instead just look it up from the infrastructure object each time? We should be able to set up an informer for the infrastructure so that we have a cache.

The reason I don't like this is because the user could override this, and it creates a weird semantic for them, why do they care what platform it's on? It really should be internal to CAO IMO

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -289,6 +293,13 @@ func (r *Reconciler) UpdateAutoscaler(ca *autoscalingv1.ClusterAutoscaler) error
 		return err
 	}
 
+	// Update the cluster provider type.
+	platformType, err := r.getPlatformType()
+	if err != nil {
+		return err
+	}
+	r.config.platformType = platformType
```
elmiko, line 301: do we need to set this on a `CreateAutoscaler` as well, or will we get an update after creating?

elmiko (commented):
generally looks good, i have a question and a suggestion.

elmiko (commented):
thanks @odvarkadaniel !

/approve
/lgtm

elmiko (approved):

enxebre:
One of the main values of CAPI is that is cloud agnostic for consumers. Moving this cloud provider detail up here is a design regression in that sense. Let's try to not leak cloud provider details as we keep moving forward.

elmiko:
in general i agree with this notion @enxebre , but it seems like the cluster-api community would prefer to not encode these platform specific values into the cluster-autoscaler itself. it seems like the CAO is the right place to be adding these platform specific flags, do you think we should be pushing harder to have these values encoded into the autoscaler itself?

PR #263, opened 2023-01-09
Add `--expander` option for cluster-autoscaler

MaXinjian:
Adds `--expander` option for cluster-autoscaler

Sets `--expander=<expander name>` CA feature flag. 

This feature provides strategies for selecting the node group to which new nodes will
be added. It is already supported in upstream [cluster-autoscaler](https://github.com/openshift/kubernetes-autoscaler/tree/master/cluster-autoscaler) . Here only expose it in operator.

5 expanders now: `random`, `most-pods`, `least-waste`, `price`, `priority`.
If leave without setting, use `random` by default.

Signed-off-by: Ma Xinjian <maxj.fnst@fujitsu.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ The operator manages the following custom resources:
       --leader-elect-lease-duration=137s
       --leader-elect-renew-deadline=107s
       --leader-elect-retry-period=26s
+      --expander=random
       --expendable-pods-priority-cutoff=-10
       --max-nodes-total=24
       --cores-total=8:128
--- a/examples/clusterautoscaler.yaml
+++ b/examples/clusterautoscaler.yaml
@@ -6,6 +6,7 @@ metadata:
 spec:
   logVerbosity: 1
   balanceSimilarNodeGroups: true
+  expander: "random"
   ignoreDaemonsetsUtilization: false
   skipNodesWithLocalStorage: true
   podPriorityThreshold: -10
--- a/install/01_clusterautoscaler.crd.yaml
+++ b/install/01_clusterautoscaler.crd.yaml
@@ -56,6 +56,12 @@ spec:
                 items:
                   type: string
                 type: array
+              expander:
+                description: Sets `--expander=<expander name>` CA feature flag. This feature provide
+                  strategies for selecting the node group to which new nodes will be added. 5
+                  expanders now: `random`, `most-pods`, `least-waste`, `price`, `priority`.
+                  `random` by default. 
+                type: string
               ignoreDaemonsetsUtilization:
                 description: Enables/Disables `--ignore-daemonsets-utilization` CA
                   feature flag. Should CA ignore DaemonSet pods when calculating resource
--- a/pkg/apis/autoscaling/v1/clusterautoscaler_types.go
+++ b/pkg/apis/autoscaling/v1/clusterautoscaler_types.go
@@ -16,6 +16,12 @@ type ClusterAutoscalerSpec struct {
 	// Configuration of scale down operation
 	ScaleDown *ScaleDownConfig `json:"scaleDown,omitempty"`
 
+	// Sets `--expander=<expander name>` CA feature flag. This feature provide
+	// strategies for selecting the node group to which new nodes will be added. 5
+	// expanders now: `random`, `most-pods`, `least-waste`, `price`, `priority`.
+	// `random` by default.
+	Expander string `json:"expander,omitempty"`
+
 	// Gives pods graceful termination time before scaling down
 	MaxPodGracePeriod *int32 `json:"maxPodGracePeriod,omitempty"`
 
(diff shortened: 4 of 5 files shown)


elmiko:
hi @MaXinjian , thanks for the contribution but we will need to discuss adding these expanders. not all of them exist for the cluster-api provider that we use, and we will need to plan for supporting the others. i am placing a hold on this PR so that our team can discuss these changes.

/hold for team review

elmiko:
@MaXinjian , i have talked with our team a little more about this PR and the larger feature to add the other expanders. i would like to keep this PR on hold for the time being as we will need to figure out a few things before we could merge this:

1. we need a feature request in our JIRA system. we did have one in the past but it was rejected. we use the feature requests to help plan the engineering, documentation, and support for new features.
2. we need to figure out which expanders we can support. as i noted above, we will not be able to use the price expander as it is not implemented for cluster-api, but we might be able to use the others. this will require some investigation.
3. we need to coordinate with docs and quality engineering to ensure we have the proper information and tests.

i hope that gives a better picture into how we are approaching this, it will take us some time to figure out the answers to these questions but i can continue to update here when we have new information.

out of curiosity, have you tested this patch on a live cluster?

if so, which expander options did you try?

MaXinjian:
> @MaXinjian , i have talked with our team a little more about this PR and the larger feature to add the other expanders. i would like to keep this PR on hold for the time being as we will need to figure out a few things before we could merge this:
> 
> 1. we need a feature request in our JIRA system. we did have one in the past but it was rejected. we use the feature requests to help plan the engineering, documentation, and support for new features.
> 2. we need to figure out which expanders we can support. as i noted above, we will not be able to use the price expander as it is not implemented for cluster-api, but we might be able to use the others. this will require some investigation.
> 3. we need to coordinate with docs and quality engineering to ensure we have the proper information and tests.
> 
> i hope that gives a better picture into how we are approaching this, it will take us some time to figure out the answers to these questions but i can continue to update here when we have new information.
> 
> out of curiosity, have you tested this patch on a live cluster?
> 
> if so, which expander options did you try?

    Thanks for your comments, and sorry for the late reply.
    Yes, we have tested this patch in a live OCP cluster, added expander in ClusterAutoscaler CR, and it can be correctly applied to the startup parameters of cluster-autoscaler.

    But we have not specifically tested which expanders it supports, because we thought that the cluster-autoscaler project has already implemented these expanders, so we thought that they are all supported. But judging from your reply, some expanders may not be supported in the cluster-api. We will try to test and figure out if they work normally in OCP, and will update here when we have results.

fenggw-fnst:
> out of curiosity, have you tested this patch on a live cluster?
if so, which expander options did you try?

Hi, I'm MaXinjian's colleague.
We have tested `most-pods`, `least-waste` and `priority` on a live cluster of 4.13.0-ec.2, they all work correctly.
To be specific, we don't have enough physical baremetal servers to do the real scale up, but according to the changes of MachineSet's replicas and Autoscaler's logs we can see that the node group calculation and scale up numbers are all as expected.

elmiko:
@fenggw-fnst thank you for reporting your findings, this will be helpful for the feature request.

fenggw-fnst:
> > out of curiosity, have you tested this patch on a live cluster?
> > if so, which expander options did you try?
> 
> Hi, I'm MaXinjian's colleague. We have tested `most-pods`, `least-waste` and `priority` on a live cluster of 4.13.0-ec.2, they all work correctly. To be specific, we don't have enough physical baremetal servers to do the real scale up, but according to the changes of MachineSet's replicas and Autoscaler's logs we can see that the node group calculation and scale up numbers are all as expected.

Sorry I forgot to mention the minor trouble when testing with priority expander.

By default, the SA of Cluster Autoscaler cannot list and watch ConfigMap, and I had to add corresponding permissions to the role used by the SA.

elmiko:
ack, thanks for the update @fenggw-fnst

guimou:
Hi! What's the status of this?
More precisely, is it possible to use the priority expander in an OpenShift cluster in a simple way? With OpenShift Data Science, we often have machine sets of various types of instances, especially with GPUs (not the same number of GPUs per node for example). So setting a priority would be nice to solve cases like: user requests 2 GPUs, which could be fulfilled with a new typeA machine that has 2 GPUs, but instead because the default is expander, it starts a machine with 8 GPUs, which is useless and expensive.
Or am  I on the wrong path here?

elmiko:
> Hi! What's the status of this?

we currently do not support changing the expander option for the autoscaler. to bring this work into openshift. as this feature will require some changes on our part, and we will need to add more support options, we need this feature to go through the feature request (RFE) process. as far as i know, this process has not been started.

> So setting a priority would be nice to solve cases like: user requests 2 GPUs, which could be fulfilled with a new typeA machine that has 2 GPUs, but instead because the default is expander, it starts a machine with 8 GPUs, which is useless and expensive.
Or am I on the wrong path here?

no, i don't think so. we are currently using the `random` expander, which means it could choose a less efficient machine by random change. it sounds like you would want the `least-waste` expander option at the minimum. i have a feeling that we could support all the options except for `price`.

elmiko:
i am going to close this issue as we have chosen to reject this feature request, see https://issues.redhat.com/browse/RFE-1294

i would encourage openshift users who want to see this feature to reach out to their red hat account manager (if available) to request the feature, with enough requests we could revisit the decision to reject. also, i would reach out in the cluster-api community to encourage adoption of this feature through the autoscaler.

currently, the cluster-api provider is what openshift uses in the cluster-autoscaler. this provider does not currently support all the expanders (for example "price"). if the upstream community is willing to implement the necessary changes for all the expanders, then it will be easier to include that functionality in openshift at some point in the future.

/close

PR #265, opened 2023-01-11
Add infrastructures resource to config.openshift.io

odvarkadaniel:
Fix for this story: https://issues.redhat.com/browse/OCPCLOUD-1702. QE was still running into some issue with the `Infrastructures` resource not being able to be listed in API group `config.openshift.io` at the cluster scope.

### DIFF
--- a/install/03_rbac.yaml
+++ b/install/03_rbac.yaml
@@ -31,6 +31,7 @@ rules:
   - get
   - update
   - list
+  - watch
 - apiGroups: ["authentication.k8s.io"]
   resources:
   - tokenreviews
--- a/install/03_rbac.yaml
+++ b/install/03_rbac.yaml
@@ -25,6 +25,7 @@ rules:
   resources:
   - clusteroperators
   - clusteroperators/status
+  - infrastructures
   verbs:
   - create
   - get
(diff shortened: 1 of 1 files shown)


odvarkadaniel:
Also had to add another verb `watch`. Was getting an error in the cluster: `Failed to watch *v1.Infrastructure: unknown (get infrastructures.config.openshift.io)`. This seems to solve it.

install/03_rbac.yaml
```diff
@@ -25,11 +25,13 @@ rules:
   resources:
   - clusteroperators
   - clusteroperators/status
+  - infrastructures
   verbs:
   - create
   - get
   - update
   - list
+  - watch
```
damdo, line 34: I would create a new entry for `infrastructures` where we only give `get/list/watch` (provided they are what we need). Granting write operations to the `infrastructures` objects is extra powers that we don't need in this case.

damdo (changes requested):

install/03_rbac.yaml
odvarkadaniel: Makes sense. Updated.

damdo (commented):
/lgtm

We can ask QE to check if this fixes it for them too. Then we can merge I think :)

JoelSpeed:
CC @sunzhaohua2 Could you please pre-merge verify this resolves the permissions issue, looks like it should to me

PR #268, opened 2023-01-26
Bug 1943194: update GPU resource limits type to have validation

elmiko:
This change set adds some behavior to the CAO to help users understand when they have poorly formed values for the ResourceLimits.GPUS.Type value, and also to help users see when their MachineSets have missing, or poorly formed, values for the GPU accelerator label. The CAO will validate CA objects and return a warning, but not deny admission, when a poorly formed value is seen. Additionally, when updating the MachineSet to add the min/max limits, if a GPU is detected and the corresponding MachineSet does not have a GPU accelerator label, a warning log and event will be emitted to help users understand the behavior. The warning messages also link to the KCS article describing the behavior.

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1943194


elmiko:
/hold

we should discuss this before merging since it will have an impact on users who currently have a `resourceLimits.GPUS.Type` field with a bad value.

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -107,6 +107,12 @@ type ResourceLimits struct {
 }
 
 type GPULimit struct {
+	// The type of GPU to associate with the minimum and maximum limits.
+	// This value is used by the Cluster Autoscaler to identify Nodes that will have GPU capacity by searching
+	// for it as a label value on the Node objects. For example, Nodes that carry the label key
+	// `cluster-api/accelerator` with the label value being the same as the Type field will be counted towards
+	// the resource limits by the Cluster Autoscaler.
+	// +kubebuilder:validation:MaxLength=63
```
JoelSpeed, line 115: Adding a maximum length to this could cause existing stored values to be invalid, that would be a breaking change.
Any validations we make to this field must be done in a way that doesn't cause issues with stored data, eg, we must warn users that their values are invalid, and maybe monitor this with a view to getting to zero warnings across our fleet, if we can prove the change wouldn't break anyone, we can make the change

I think this is going to need to be a controller based validation with good error reporting back

elmiko: yeah, even aside from the max length, the label value validation will also cause some failures as we have been telling folks to put things like `nvidia.com/gpu` in that field.

elmiko: we talked about this today, i am going to rework this patch so that we produce warnings for validating the `GPUS.Type` field. we will also emit a log message and event when we find a machineset that has gpu capacity but no associated `.spec.template.spec.metadata.labels` for the accelerator label.

in this way we won't break current users, but still provide good information about how to fix. i will also include a link to [this KCS article](https://access.redhat.com/solutions/6055181) about the underlying issue.

elmiko:
ok, i've totally refactored this PR to do validation on the GPU type values for the ClusterAutoscaler resource, and to emit warnings when a poorly formatted accelerator label is found. i am updating the description and removing the hold.

/cancel hold

elmiko:
@JoelSpeed the one thing i wasn't quite sure about was if we want to add a test to detect that the event is being emitted properly.

pkg/controller/machineautoscaler/machinetarget_test.go
```diff
@@ -283,3 +283,70 @@ func TestNamespacedName(t *testing.T) {
...
 	}
 }
+
+func TestHasGPUCapacity(t *testing.T) {
+	target := NewTarget()
+	target.SetAnnotations(map[string]string{
+		autoscalerCapacityGPU: "1",
+	})
+
+	if !target.HasGPUCapacity() {
```
JoelSpeed, line 293: Should each of these cases be a subtest?

pkg/util/validator.go
```diff
@@ -0,0 +1,59 @@
...
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+const (
+	GPUAcceleratorLabelKCSWarning string = "For more information on the proper use of these values, please see https://access.redhat.com/solutions/6055181"
+)
+
+type ValidatorResponse struct {
```
JoelSpeed, line 30: Missing godoc

pkg/util/validator.go
```diff
@@ -0,0 +1,59 @@
...
+	Warnings []string
+	Errors   utilerrors.Aggregate
+}
+
+func (vr ValidatorResponse) IsValid() bool {
```
JoelSpeed, line 35: Missing godoc

JoelSpeed (commented):
Some unit tests around the various messages being emitted in the events would be good, what did you have in mind?

pkg/controller/machineautoscaler/machinetarget_test.go
elmiko: yeah, agreed

elmiko:
i tested this on a live cluster but it wasn't quite behaving how i expect, didn't see log messages, events, or warning on the webhook creation. need to investigate more.

elmiko:
got this working in cluster today, had a couple bugs that needed fixing. im also updating to address the comments from the review.

elmiko:
@JoelSpeed i'm fairly confident in this PR, but happy to have a qe test. should i bring it up on the bug?

JoelSpeed:
Yes, if they can pre-merge verify that would be good, not sure how feasible that is given the limitations in testing the NFD stuff, is that still needed?

elmiko:
@miyadav did some testing to ensure that the warnings and events are being properly created. i am going to remove the hold here, i think we should merge this and then retest the full pipeline to ensure that the limits work properly as well.

/hold cancel

PR #269, opened 2023-02-03
CNF-5642 make operator optional

qJkee:
Add capability annotation to manifests to make operator optional


qJkee:
on hold until https://github.com/openshift/api/pull/1292 is merged

JoelSpeed:
Seems the tooling in this repo doesn't like additional annotations being added to CRDs randomly, you'll need to check what's happening there and work out a way to add additional annotations

qJkee:
@JoelSpeed 
Fixed!

JoelSpeed:
Looking at the test results in the cluster, this appears to make the cluster autoscaler manifests not deploy, is that intentional? 
We cannot suddenly stop deploying the autoscaler operator as that would break upgrades, what's the plan here?

PR #274, opened 2023-05-23
OCPCLOUD-2061: Rebase dependencies to k8s 1.27

elmiko:
also update the controller-runtime dependency and fix some issues related to the changes.


JoelSpeed:
/lgtm cancel

Panics in the logs, https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-autoscaler-operator/274/pull-ci-openshift-cluster-autoscaler-operator-master-e2e-aws-operator/1663865391102824448/artifacts/e2e-aws-operator/gather-extra/artifacts/pods/openshift-machine-api_cluster-autoscaler-operator-647759f49d-vhbh2_cluster-autoscaler-operator.log

elmiko:
looking at the output, my best guess is that this is related to the controller-runtime changes.

elmiko:
> Looks like the panic is coming because the logger is nil, can we set up a logger easily?

i hope so

JoelSpeed:
[InjectDecoder](https://github.com/openshift/cluster-autoscaler-operator/blob/f7f168cc5145da8b8c781122a39582e1a2ad39ce/pkg/controller/clusterautoscaler/validator.go#L205-L209) is no longer being called as of v0.15.0 of controller runtime (the client one probably isn't either?) so we need to update the constructor of the validator to avoid using these inject style methods, the code is panicking in the last instance because the decoder is nil

elmiko:
@JoelSpeed ptal, it seems like we are back to normal autoscaler failures, i think CAO is happy now.

pkg/controller/clusterautoscaler/validator.go
```diff
@@ -27,8 +28,10 @@ type Validator struct {
 
 // NewValidator returns a new Validator configured with the given
 // ClusterAutoscaler singleton resource name.
-func NewValidator(name string) *Validator {
+func NewValidator(name string, client client.Client, scheme *runtime.Scheme) *Validator {
```
JoelSpeed, line 31: Should we drop the `InjectDecoder` and `InjectClient` functions now then?

pkg/controller/machineautoscaler/validator.go
```diff
@@ -20,8 +21,11 @@ type Validator struct {
 }
 
 // NewValidator returns a new Validator.
-func NewValidator() *Validator {
-	return &Validator{}
+func NewValidator(client client.Client, scheme *runtime.Scheme) *Validator {
```
JoelSpeed, line 24: Same again, should we drop the inject functions?

dtobolik:
Regression looks good on this one, ping me for any major changes
/label qe-approved

PR #275, opened 2023-06-07
OCPBUGS-14356: add nutanix labels that should be ignored

elmiko:
This change ammends the balancing label ignore logic to include the Nutanix specific labels which should be ignored.

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler.go
@@ -72,6 +72,12 @@ const (
 // The following values are for cloud providers which have not yet created specific nodegroupset processors.
 // These values should be removed and replaced in the event that one of the cloud providers creates a nodegroupset processor.
 
+// Alibaba cloud provider ignore labels for the autoscaler.
+const (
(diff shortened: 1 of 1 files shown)


thunderboltsid:
/retest-required
LGTM

JoelSpeed:
Is there a periodic nutanix job in the payload test jobs that we could use to test this?

elmiko:
i'm not sure, i'll have to do some investigation on that

elmiko:
@JoelSpeed i found a couple different ways we could use the nutanix jobs to test this, the best is probably going to be running our cluster-api-actuator-pkg suite against a nutanix cluster.

is there a convenient way to run it and show the results here without adding that job to the prow config? (i notice we only have aws jobs here currently)

JoelSpeed:
We have to set it up on the prow config, but we don't have to have it set to always run, we can make it on demand only

elmiko:
ack, i'll take a look into adding it as optional here.

elmiko:
let's see if this works

/test e2e-nutanix-operator

elmiko:
@JoelSpeed looks like we passed the nutanix test

PR #277, opened 2023-06-08
WIP: decrease scan interval to 5 seconds

elmiko:
doing some testing in conjunction with https://github.com/openshift/kubernetes-autoscaler/pull/258


elmiko:
the autoscaler gets really unhappy with a scan-interval of 5s

PR #279, opened 2023-07-26
Reconcile when external process change the clusteroperator status

mfojtik:
Suppose an external entity (admin, script, automation...) changes the message or reason in the cluster operator instance of this operator. In that case, we need a way to detect this and stomp it back.
This change wire an informer that watches for changes in cluster operators and calls the report status if a change is detected.


pkg/operator/status.go
```diff
@@ -295,8 +302,29 @@ func (r *StatusReporter) progressing(reason, message string) error {
...
+	// run informer to make sure that we report status everytime the cluster operator changes
+	// for example. if some external process override message or reason.
+	go r.configInformers.Start(stop.Done())
+	operatorInformer := r.configInformers.Config().V1().ClusterOperators().Informer()
+	if !cache.WaitForCacheSync(stop.Done(), operatorInformer.HasSynced) {
+		return fmt.Errorf("unable to sync clusteroperators informer")
+	}
+	_, err := operatorInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			r.ReportStatus()
```
mfojtik, line 314: normally this would just enqueue a key, not sure how to do it in controller-runtime

JoelSpeed: This controller isn't using the normal patterns in controller-runtime, in fact, it has no queue and isn't currently using any events to watch. Instead, when this controller is started, it does a poll and just ticks to update the status every 15 seconds.

The ideal would be to refactor this to use the proper controller-runtime patterns, which would mean it could watch the clusteroperator, and tick relatively easily, but, that's a little harder to do.

Is this working like this? Or does it need to be updated?

mfojtik: @JoelSpeed I added the queue just now, otherwise I would feel horrible :-)

mfojtik:
/hold cancel

I reworked this to use a standard work queue when triggering status updates;

sanchezl (approved):
/lgtm

PR #284, opened 2023-09-14
OCPBUGS-18137: Provide the architecture of the control plane as argument to --scale-up-from-zero-default-arch

aleskandro:
Depends on https://github.com/openshift/kubernetes-autoscaler/pull/261

When autoscale from zero is enabled in non-amd64 single-arch clusters on cloud providers managed by actuators that do not set the architecture in the capacity annotations (ref. [MIXEDARCH-129](https://issues.redhat.com//browse/MIXEDARCH-129)), passing the argument --scale-up-from-zero-default-arch will allow not to default on the amd64 architecture for building the generic labels to scale out nodegroups from zero.

This commit sets the argument for the cluster-autoscaler to runtime.GOARCH.

/hold

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler.go
@@ -2,6 +2,7 @@ package clusterautoscaler
 
 import (
 	"fmt"
+	"runtime"
 
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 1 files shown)


elmiko (commented):
/lgtm

i'd like to talk with the team a little about the autoscaler carry patch, but i don't imagine we'll have pusback.

PR #285, opened 2023-09-14
OCPBUGS-18954: Ensure status reporter caches exit if they don't sync

JoelSpeed:
We are seeing sometimes, that the Cluster Autoscaler status never gets reported. Looking at the logic, it's possible here that we wait forever for the informers to sync, but, if they never actually sync, the status won't get reported.
I'm not sure why they wouldn't sync, but, having a timeout seems reasonable to prevent this infinite hang.

If we start seeing this timeout hit repeatedly, then my theory is correct, and we can then investigate why the informers aren't syncing correctly.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -338,12 +338,18 @@ func (c *StatusReporter) runWorker(queueCtx context.Context) {
 func (r *StatusReporter) Start(stop context.Context) error {
 	// run informer to make sure that we report status everytime the cluster operator changes
 	// for example. if some external process override message or reason.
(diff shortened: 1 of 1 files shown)


elmiko (commented):
makes sense to me, i have a question about one of your comments though.

> If we start seeing this timeout hit repeatedly,

how will we know?  i'm just curious what the signal from ci will look like.

/lgtm
/approve

JoelSpeed:
> how will we know? i'm just curious what the signal from ci will look like.

CI has a check already that will report if a pod has restarted many times, so, if it becomes an issue and the pod is restarting every 30 seconds, the existing suite should tell us

PR #294, opened 2023-10-10
OCPCLOUD-2137: Check scale from zero annotations on MachineSets

odvarkadaniel:
Currently, we support only `machine.openshift.io` annotations for scaling from zero - this PR introduced logic that let's the user use the upstream prefix `capacity.cluster-autoscaler.kubernetes.io` for this purpose. We also need to make sure to correctly calculate between `memoryMb` and `memory` suffixes for their corresponding prefixes. We want to have both sets of these annotation present on the MachineSet.


odvarkadaniel:
cc @elmiko for review

pkg/controller/machineautoscaler/machinetarget.go
```diff
@@ -280,3 +291,67 @@ func (mt *MachineTarget) WarningForInvalidGPUAcceleratorLabel() string {
...
+	annotations := mt.GetAnnotations()
+
+	if annotations == nil {
+		return nil
+	}
+
+	err := checkScaleFromZeroAnnotations(annotations)
+	if err != nil {
+		// canScale, err = checkNewScaleFromZeroAnnotations(annotations)
+		// if !canScale && err != nil {
```
elmiko, line 309: are these comments left over?

pkg/controller/machineautoscaler/machinetarget.go
```diff
@@ -280,3 +291,67 @@ func (mt *MachineTarget) WarningForInvalidGPUAcceleratorLabel() string {
...
+			return fmt.Errorf("could not parse integer: %w", err)
+		}
+		annotations[memoryKey] = resource.NewQuantity(memInt*util.MiB, resource.DecimalSI).String()
+	}
+
+	// Original GPU annotation is split into 2 upstream annotations
+	gpu := parseKey(annotations, gpuCountKeyOriginal)
+	if gpu != "" {
+		annotations[gpuCountKey] = gpu
+		annotations[gpuTypeKey] = "nvidia.com/gpu"
```
elmiko, line 340: could you add a `// TODO ` comment here saying that we need to come back and fix this once we have the proper gpu types.

elmiko (commented):
this is looking good Daniel, i like the comments too. just a couple comments.

pkg/controller/machineautoscaler/machinetarget.go
odvarkadaniel: Added.

odvarkadaniel:
> this is looking good Daniel, i like the comments too. just a couple comments.

Addressed your comments, Mike. Should be ready to review.

odvarkadaniel:
The CI is now expected to fail until the module work is merged in MAO (https://github.com/openshift/machine-api-operator/pull/1169).

pkg/controller/machineautoscaler/machinetarget.go
```diff
@@ -280,3 +281,65 @@ func (mt *MachineTarget) WarningForInvalidGPUAcceleratorLabel() string {
...
+	// Deprecated GPU annotation is split into 2 upstream annotations.
+	// TODO: Once we introduce proper gpu types, the gpu type in the annotation module in MAO needs to be changed.
+	gpu, err := annotationsutil.ParseMachineSetAnnotationKey(annotations, annotationsutil.GpuCountKeyDeprecated)
+	if err != nil {
+		return fmt.Errorf("could not parse key %s: %w", annotationsutil.GpuCountKeyDeprecated, err)
+	}
+	annotations = annotationsutil.SetGpuCountAnnotation(annotations, gpu)
+	// TODO: Once we introduce proper gpu types, the function SetGpuTypeAnnotation
+	// should take in the correct value as a second argument.
+	annotations = annotationsutil.SetGpuTypeAnnotation(annotations, "")
```
elmiko, line 336: i know we don't do anything with the second value, but let's pass  the nvidia value anyways just to make it clear what is happening.

we probably should not have defaulted the library function, but i think this is a minor detail.

elmiko (commented):
this is looking good to me, i just have one minor request

pkg/controller/machineautoscaler/machinetarget.go
odvarkadaniel: Agree. Changing the value we pass into the function.

odvarkadaniel:
> this is looking good to me, i just have one minor request

Fixed. Should be good to go.

PR #297, opened 2023-10-24
MIXEDARCH-429: Add the environment variable value for CAPI_SCALE_ZERO_DEFAULT_ARCH

aleskandro:
kubernetes/autoscaler#6066 merged using an env var instead of the command line argument, we decided to go with downstream in openshift/kubernetes-autoscaler#261. We will need to be ready for onboarding that change during the next rebase. 

This PR adds the env var to the deployment of the autscaler: we could merge this and prepare another revert PR for 6b7a17780f92775d4e08b9a594b7f1a29a75f7e8 or do both of them at once, during the next rebase. Wdyt @elmiko ?

Refers #284

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
@@ -3,6 +3,7 @@ package clusterautoscaler
 import (
 	"context"
 	"fmt"
+	goruntime "runtime"
 
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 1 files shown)


elmiko:
hey @aleskandro , apologies i missed this. i think we can merge this as it won't have negative affect on an autoscaler that doesn't know about the environment variable.

is this still good to merge?

aleskandro:
Hi, yes, I'm ok to merge this. Let me know if you prefer rebasing first

elmiko (commented):
/approve
/lgtm

PR #304, opened 2023-12-14
OCPCLOUD-2523: add ExpanderOrderList api option

elmiko:
this change allows users to set a list of expanders through this api field.


elmiko:
i'm working on improving the unit tests before proposing this, we let the arguments tests atrophy a little and i would like to improve them.

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -53,6 +64,16 @@ type ClusterAutoscalerSpec struct {
...
+
+	// Sets the type and order of expanders to be used during scale out operations.
+	// This option specifies an ordered list, highest priority first, of expanders that
+	// will be used by the cluster autoscaler to select node groups for expansion
+	// when scaling out.
+	// If not specified, the default value is `random`, available options are: `least-waste`, `priority`, `random`.
+	//
+	// +listType=set
+	// +kubebuilder:validation:MaxItems=3
+	ExpanderOrderList []ExpanderString `json:"expanderOrderList"`
```
JoelSpeed, line 76: I would probably just call this
```suggestion
	Expanders []ExpanderString `json:"exapnders"`
```

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -53,6 +64,16 @@ type ClusterAutoscalerSpec struct {
...
 	LogVerbosity *int32 `json:"logVerbosity,omitempty"`
+
+	// Sets the type and order of expanders to be used during scale out operations.
+	// This option specifies an ordered list, highest priority first, of expanders that
+	// will be used by the cluster autoscaler to select node groups for expansion
+	// when scaling out.
+	// If not specified, the default value is `random`, available options are: `least-waste`, `priority`, `random`.
+	//
+	// +listType=set
+	// +kubebuilder:validation:MaxItems=3
```
JoelSpeed, line 75: Field needs to be marked optional
```suggestion
	// +kubebuilder:validation:MaxItems=3
	// +optional
```

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -8,6 +8,17 @@ func init() {
 	SchemeBuilder.Register(&ClusterAutoscaler{}, &ClusterAutoscalerList{})
 }
 
+// ExpanderString contains the name of an expander to be used by the cluster autoscaler.
+// +kubebuilder:validation:Enum=least-waste;priority;random
```
JoelSpeed, line 12: Enums should always use PascalCase and convert inside the operator to the upstream equivalent.
```suggestion
// +kubebuilder:validation:Enum=LeastWaste;Priority;Random
```

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -8,6 +8,17 @@ func init() {
...
+type ExpanderString string
+
+// These constants define the valid values for an ExpanderString
+const (
+	LeastWasteExpander ExpanderString = "least-waste"
+	PriorityExpander   ExpanderString = "priority"
+	RandomExpander     ExpanderString = "random"
```
JoelSpeed, line 19: Use PascalCase

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -53,6 +64,16 @@ type ClusterAutoscalerSpec struct {
 	// This option has priority over log level set by the `CLUSTER_AUTOSCALER_VERBOSITY` environment variable.
 	// +kubebuilder:validation:Minimum=0
 	LogVerbosity *int32 `json:"logVerbosity,omitempty"`
+
+	// Sets the type and order of expanders to be used during scale out operations.
+	// This option specifies an ordered list, highest priority first, of expanders that
+	// will be used by the cluster autoscaler to select node groups for expansion
+	// when scaling out.
+	// If not specified, the default value is `random`, available options are: `least-waste`, `priority`, `random`.
```
JoelSpeed, line 72: Would be nice to explain what each one of those options means, and also, why you can select multiple. In which scenario would the CA fallback to the second or third expander in the list

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
elmiko: i thought that might not be descriptive enough, but if you are happy with that i'll change it.

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
elmiko: i can add that

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
elmiko: i thought that might not be descriptive enough, but if you are happy with that i'll change it.

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
elmiko: is it an error that we haven't marked the other fields as optional?

JoelSpeed: The other fields have omitempty which implicitly makes them optional, except we want to be explicit in all cases really

elmiko:
updated with suggestions

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -53,6 +64,26 @@ type ClusterAutoscalerSpec struct {
...
 	// +kubebuilder:validation:Minimum=0
 	LogVerbosity *int32 `json:"logVerbosity,omitempty"`
+
+	// Sets the type and order of expanders to be used during scale out operations.
+	// This option specifies an ordered list, highest priority first, of expanders that
+	// will be used by the cluster autoscaler to select node groups for expansion
+	// when scaling out.
+	// Expanders instruct the autoscaler on how to choose node groups when scaling out
+	// the cluster. They can be specified in order so that the autoscaler will utilize
+	// the result from one expander on the next, for example if the first expander option
```
JoelSpeed, line 74:
```suggestion
	// the result from one expander or the next, for example if the first expander option
```

JoelSpeed (commented):
One typo I think but otherwise LGTM

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
elmiko: that is actually correct, this is meant to tell people that the expanders work like filters.

if the user specifies `expanders: [LeastWaste, Random]`, the autoscaler will first determine which node group(s) will satisfy least waste, if there are multiple groups selected, then the random expander will choose one of those.

elmiko: maybe it would be clearer if i said this?

> They can be specified in order so that the result from the first expander is used as input to the second, and so forth.

JoelSpeed: Yeah I think that would be a clearer way to word it

elmiko:
updated the api description to be clearer.

PR #309, opened 2023-12-31
ExtraArgs can take multiple parameters

ikukaku:
ExtraArgs can use multiple parameters, for example: "--max-empty-bulk-delete=50,--scale-down-utilization-threshold=0.5"

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	goruntime "runtime"
+	"strings"
 
 	configv1 "github.com/openshift/api/config/v1"
 	autoscalingv1 "github.com/openshift/cluster-autoscaler-operator/pkg/apis/autoscaling/v1"
@@ -416,7 +417,7 @@ func (r *Reconciler) AutoscalerPodSpec(ca *autoscalingv1.ClusterAutoscaler) *cor
 	args := AutoscalerArgs(ca, &r.config)
 
 	if r.config.ExtraArgs != "" {
-		args = append(args, r.config.ExtraArgs)
+		args = append(args, strings.Split(r.config.ExtraArgs, ",")...)
 	}
 
 	spec := &corev1.PodSpec{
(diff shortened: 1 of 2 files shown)


elmiko:
thanks @ikukaku , this looks like a nice upgrade. i have one request, would you mind writing a unit test to exercise this new option, it could look similar to [this test for enabled arguments](https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/controller/clusterautoscaler/clusterautoscaler_test.go#L192-L216).

ikukaku:
> --max-empty-bulk-delete=50,--scale-down-utilization-threshold=0.5



> thanks @ikukaku , this looks like a nice upgrade. i have one request, would you mind writing a unit test to exercise this new option, ~it could look similar to [this test for enabled arguments](https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/controller/clusterautoscaler/clusterautoscaler_test.go#L192-L216).~
> 
> edit: apologies, i looked at the implementation again and it seems we would need to test this from the reconcile unit test. i would love to get this in a test, but it might be tough to test because of the way that `ExtraArgs` is injected through the environment variable.

thank for your reply.

it is hard to get deployment args in reconcile  unit  test. I would prefer to add ut for   [`AutoscalerPodSpec`](https://github.com/ikukaku/cluster-autoscaler-operator/blob/efb05803c9bbc48f9b79c1e56c040359ceea743d/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go#L414). is that ok?

ikukaku:
> thanks @ikukaku , this looks like a nice upgrade. i have one request, would you mind writing a unit test to exercise this new option, ~it could look similar to [this test for enabled arguments](https://github.com/openshift/cluster-autoscaler-operator/blob/master/pkg/controller/clusterautoscaler/clusterautoscaler_test.go#L192-L216).~
> 
> edit: apologies, i looked at the implementation again and it seems we would need to test this from the reconcile unit test. i would love to get this in a test, but it might be tough to test because of the way that `ExtraArgs` is injected through the environment variable.

@elmiko  thanks for your reply.

 ut for `multi-extraArgs` is done. Looking forward to more feedback.

elmiko:
unit test look great, thanks for updating @ikukaku !

i just want to make sure everything looks good in ci, then happy to approve this.

ikukaku:
@elmiko  Hello, is there anything else that needs to be done with this PR?

elmiko:
apologies @ikukaku , happy to approve.

/approve
/lgtm

elmiko:
not sure what happened here, i think we need to generate a jira card for this. i will followup. apologies for the confusion @ikukaku 

 /remove-lifecycle stale

PR #316, opened 2024-03-04
OCPCLOUD-2493: Update CAO to add upstream annotations

racheljpg:
Hello! This is a PR to update the CAO to add the upstream scale from zero annotations if it's only the old annotations that are there. Thanks!

### DIFF
--- a/pkg/controller/machineautoscaler/machinetarget_test.go
+++ b/pkg/controller/machineautoscaler/machinetarget_test.go
@@ -2,9 +2,11 @@ package machineautoscaler
 
 import (
 	"fmt"
+	"maps"
 	"testing"
 
 	"github.com/openshift/cluster-autoscaler-operator/pkg/util"
+	annotationsutil "github.com/openshift/machine-api-operator/pkg/util/machineset"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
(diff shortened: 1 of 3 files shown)


elmiko (commented):
changes are looking good, we need some cleanup and more tests

pkg/controller/machineautoscaler/machinetarget_test.go
```diff
@@ -375,3 +377,100 @@ func TestWarningForInvalidGPUAcceleratorLabel(t *testing.T) {
...
+
+			if !maps.Equal(observed, tc.expectedNewAnnotation) {
+				t.Errorf("SetAnnotations() returned %v, expected %v", observed, tc.expectedNewAnnotation)
+			}
+		})
+	}
+}
+
+// test checkscalefromzeroannotations on a new machine
+func TestScaleFromZeroAnnotations(t *testing.T) {
```
racheljpg, line 452: @elmiko Do you think it's worth adding this as a test case above to make things neater, or is it fine to leave it out here on it's own?

pkg/controller/machineautoscaler/machinetarget_test.go
```diff
@@ -375,3 +377,77 @@ func TestWarningForInvalidGPUAcceleratorLabel(t *testing.T) {
...
+			name: "Supplying old CPU annotation and adding new one",
+			expectedNewAnnotation: map[string]string{
+				annotationsutil.CpuKey:           "1",
+				annotationsutil.CpuKeyDeprecated: "1",
+			},
+			suppliedAnnotations: map[string]string{
+				annotationsutil.CpuKeyDeprecated: "1",
+			},
+		},
+	}
```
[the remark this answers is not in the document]
racheljpg: Hey Mike, thank you for the feedback. I've added some more test cases to test the other annotations in the SetAnnotations function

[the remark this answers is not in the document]
racheljpg: Also added a second commit to make some changes to commits/names for clarity!

pkg/controller/machineautoscaler/machinetarget_test.go
```diff
@@ -375,3 +377,100 @@ func TestWarningForInvalidGPUAcceleratorLabel(t *testing.T) {
...
+			suppliedAnnotations: map[string]string{
+				annotationsutil.CpuKeyDeprecated: "1",
+			},
+		},
+		{
+			name: "Supplying old GPU annotation and adding upstream annotations",
+			expectedNewAnnotation: map[string]string{
+				annotationsutil.GpuCountKey:           "1",
+				annotationsutil.GpuCountKeyDeprecated: "1",
+				annotationsutil.GpuTypeKey:            "1",
```
elmiko, line 414: just to make this a little more realistic, i would change this value to `nvidia.com/gpu`, that is how it will look in a live cluster.

pkg/controller/machineautoscaler/machinetarget_test.go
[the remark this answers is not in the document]
elmiko: these tests are looking great @racheljpg !

pkg/controller/machineautoscaler/machinetarget_test.go
elmiko: this seems like we are testing the same functionality as above, i would add it as a case there.

elmiko (commented):
apologies for the delay on getting back here, this is looking good to me, just a couple minor nits and we can probably merge this.

pkg/controller/machineautoscaler/machinetarget_test.go
racheljpg: It actually seems like we are doing what I was trying to test there above on line 389, so I've removed that function entirely :)

racheljpg:
Hello @elmiko, no worries, and thank you. I've changed the value of the GpuTypeKey and removed a redundant test. Thanks!

racheljpg:
Hello!
I've tested this manually (thanks @elmiko) and can confirm that the new annotations show up on a MachineSet. :)

```apiVersion: machine.openshift.io/v1beta1
kind: MachineSet
metadata:
  annotations:
    machine.openshift.io/GPU: '0'
    autoscaling.openshift.io/machineautoscaler: openshift-machine-api/worker-us-east-1a
    machine.openshift.io/memoryMb: '16384'
    capacity.cluster-autoscaler.kubernetes.io/gpu-type: nvidia.com/gpu
    capacity.cluster-autoscaler.kubernetes.io/labels: kubernetes.io/arch=amd64
    machine.openshift.io/cluster-api-autoscaler-node-group-min-size: '1'
    capacity.cluster-autoscaler.kubernetes.io/memory: '17179869184'
    capacity.cluster-autoscaler.kubernetes.io/cpu: '4'
    capacity.cluster-autoscaler.kubernetes.io/gpu-count: '0'
    machine.openshift.io/vCPU: '4'
    machine.openshift.io/cluster-api-autoscaler-node-group-max-size: '12'```

elmiko (commented):
/approve
/lgtm

PR #317, opened 2024-04-10
OCPCLOUD-2493: Add empty string as value for GpuType if GpuCount is empty 0

racheljpg:
Hello, 
This is a bug fix for [OCPCLOUD-2493](https://issues.redhat.com/browse/OCPCLOUD-2493). 
The GpuType annotation needs to be empty if the GpuCount annotation = 0.
This PR also revendors the MAO so that it can bring in the change [in this PR](https://github.com/openshift/machine-api-operator/pull/1226), removing a default value we no longer want.
Thanks!


pkg/controller/machineautoscaler/machinetarget.go
```diff
@@ -341,7 +341,12 @@ func checkScaleFromZeroAnnotations(annotations map[string]string) (map[string]st
 		}
 		_, err = annotationsutil.ParseMachineSetAnnotationKey(annotations, annotationsutil.GpuTypeKey)
 		if err != nil {
-			annotations = annotationsutil.SetGpuTypeAnnotation(annotations, annotationsutil.GpuNvidiaType)
+			//If there are no gpus, gputype needs to be an empty string
```
elmiko, line 344: we need to make sure that the type label is applied even when it did not exist previously. i think you need to double check the logic for how we enter these clauses.

we want to add the type label, and we currently gate this logic on whether it exists already.

elmiko: ok, so it seems that the `SetGpuTypeAnnotation` is the culprit here.

racheljpg:
@elmiko since this has been tested and shown to work, I dropped the commit I used to test it and it should be ready for review now. Thanks!

elmiko (commented):
this is looking good, i think we should add a unit test case to make sure we get the empty string when it's set to "0".

racheljpg:
Closing this as this case is a bit redundant (with another case already testing this specific setter) and to move this logic to the CAO (ref Mike's comment above).

pkg/controller/machineautoscaler/machinetarget_test.go
```diff
@@ -418,6 +418,18 @@ func TestUpdatingScaleFromZeroAnnotations(t *testing.T) {
...
+		{
+			name: "Supplying old GPU annotations and adding upstream annotations with empty string if count is 0",
+			expectedNewAnnotation: map[string]string{
+				annotationsutil.GpuCountKey:           "0",
+				annotationsutil.GpuCountKeyDeprecated: "0",
+				annotationsutil.GpuTypeKey:            "",
+			},
+			suppliedAnnotations: map[string]string{
+				annotationsutil.GpuCountKeyDeprecated: "0",
+				annotationsutil.GpuTypeKey:            "",
```
elmiko, line 430: i think we should remove this from the supplied since it wouldn't be there.

elmiko (commented):
just one change, otherwise lgtm

pkg/controller/machineautoscaler/machinetarget_test.go
racheljpg: Thanks Mike, good catch. I removed it from the case above for the same reason

PR #320, opened 2024-05-03
OCPBUGS-20565: Set webhook minimum TLS version to 1.3

theobarberbany:
This change updates the webhook server to use TLSv1.3.

Other TLS configuration options are possible, by specifying additional options in the function  defined - or defining another function.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -1,6 +1,7 @@
 package operator
 
 import (
+	"crypto/tls"
 	"fmt"
 
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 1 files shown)


pkg/operator/operator.go
```diff
@@ -65,6 +66,15 @@ func New(cfg *Config) (*Operator, error) {
 		Metrics: server.Options{
 			BindAddress: fmt.Sprintf("127.0.0.1:%d", cfg.MetricsPort),
 		},
+		WebhookServer: &webhook.DefaultServer{
+			Options: webhook.Options{
+				TLSOpts: []func(*tls.Config){
+					func(cfg *tls.Config) {
+						cfg.MinVersion = uint16(13)
```
JoelSpeed, line 73: Is there a constant we can use for this?

theobarberbany:
(Picking up from previous comment @JoelSpeed 

We're currently building with go 1.21, for which the default TLS version (acting as a server) is 1.0. 

(from the tls docs) 
> By default, TLS 1.2 is currently used as the minimum when acting as a
	// client, and TLS 1.0 when acting as a server. TLS 1.0 is the minimum
	// supported by this package, both as a client and as a server.

Once we start building with 1.22, this becomes 1.2 by default. However, this does mean we need to have this change backported to 4.14, and 4.15.

damdo (commented):
/lgtm

PR #321, opened 2024-05-03
OCPBUGS-20565: Sets webhook minimum TLS version to 1.3

theobarberbany:


pkg/operator/operator.go
```diff
@@ -57,6 +58,15 @@ func New(cfg *Config) (*Operator, error) {
 		RenewDeadline:                 &le.RenewDeadline.Duration,
 		RetryPeriod:                   &le.RetryPeriod.Duration,
 		MetricsBindAddress:            fmt.Sprintf("127.0.0.1:%d", cfg.MetricsPort),
+		WebhookServer: &webhook.DefaultServer{
+			Options: webhook.Options{
+				TLSOpts: []func(*tls.Config){
+					func(cfg *tls.Config) {
+						cfg.MinVersion = uint16(13)
```
JoelSpeed, line 65: Is there a const we can use for this?

JoelSpeed (commented):
This should be fixed first in master, and then backported, do we have a PR for master?

theobarberbany:
Yep, the other PR is for master. I just did this to see if the same fix would work for 4.14 (e2e) as the library versions differ.

I also think that this likely isn't a problem (need to double check) for master at the moment as the default version of TLS is bumped.

PR #333, opened 2024-09-30
OCPCLOUD-2739: Bump deps to 1.31

theobarberbany:


Makefile
```diff
@@ -10,7 +10,7 @@ IMAGE        = origin-cluster-autoscaler-operator
 BUILD_IMAGE ?= registry.ci.openshift.org/openshift/release:golang-1.20
 
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.29
+ENVTEST_K8S_VERSION = 1.31.0
```
JoelSpeed, line 13: Do we need to remove anything from the envtest command?

theobarberbany: wait a second i bumped this - why isn't it in the PR 😕

theobarberbany: AFAIK no, 

We define the envtest command without the extra args: 

```
ENVTEST = go run ${PROJECT_DIR}/vendor/sigs.k8s.io/controller-runtime/tools/setup-envtest
```

JoelSpeed: Invocation is  but https://github.com/openshift/cluster-autoscaler-operator/blob/546bd442b4cce7422fbaba21e90752f854c6f5a2/Makefile#L91 but that is also not specifying anything we need to remove

PR #334, opened 2024-10-23
OCPBUGS-42132: set max soft bulk taint count to zero

elmiko:
This change disables the maximum number of nodes that can be soft tainted or untainted at once for deletion.


elmiko:
even though the latest e2e-aws-periodic-pre test failed, it appears that the deletion test passed. i think we want to seriously consider removing the max bulk soft taint, but i'd like to test more.

/test e2e-aws-periodic-pre

JoelSpeed:
This may slow down bulk scale down for large clusters, but, this is preferable to the existing issues we have in the related bug

I suspect this is a case where the fix will be beneficial for the 99%, and slightly detrimental to the 1%

/approve

PR #339, opened 2025-02-06
OCPCLOUD-2829: Bump k8s 1.32

racheljpg:
Hello! This is a PR to bump the dependencies to 1.32. Thanks!

PR #340, opened 2025-02-18
WIP: Add provisioning request

elmiko:
this PR is based on top of #339 and is meant to demonstrate how we can use ProvisioningRequests with other components in openshift.


elmiko:
i'm not sure why, but it seems like the e2e-aws-periodic-pre is having some issues with the autoscaler tests. i don't think it's related to this change, but i'll need to investigate further.

elmiko:
i think i will need to test this manually to see what is happening. it appears that we are having issues with timeout on scale out.

elmiko:
i am updating this to use a feature gate, the new version will require https://github.com/openshift/api/pull/2587

jkyros:
Your infrastructure is cursed today :cry: 
/test e2e-upgrade

elmiko:
i've been manually testing this today and i think it's good to go. i can confirm that the operator deploys with the fg enabled, and it has the proper permissions, also the CAS is able to see provisioning requests.

elmiko:
i have tested this PR manually in several configurations. i am confident that it does not affect OCP without the DevPreviewNoUpgrade feature set, and i can see all the components happy when it is enabled.

/verified by elmiko

PR #343, opened 2025-06-03
OCPBUGS-57041: update cluster provider type early

maxcao13:
Fixes: https://issues.redhat.com/browse/OCPBUGS-57041

This bug was happening because the platform type of the cluster was not updated before CreateAutoscaler is being called. This results in some arguments not getting added to the deployment spec initially, but during a subsequent quick reconcile, the platform type is updated in the UpdateAutoscaler function. So immediately, the args are now added to the deployment, which triggers a new pod(s) rollout.

### DIFF
--- a/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
+++ b/pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
@@ -217,6 +217,15 @@ func (r *Reconciler) Reconcile(_ context.Context, request reconcile.Request) (re
 		return reconcile.Result{}, nil
 	}
 
+	// Update the cluster provider type.
+	if r.config.platformType == "" {
+		platformType, err := r.getPlatformType()
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		r.config.platformType = platformType
+	}
+
 	if err := r.ensureAutoscalerMonitoring(ca); err != nil {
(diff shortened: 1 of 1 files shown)


maxcao13:
/hold

Trying to use cluster-bot to run rehearsals.

pkg/controller/clusterautoscaler/clusterautoscaler_controller.go
```diff
@@ -226,6 +226,15 @@ func (r *Reconciler) Reconcile(_ context.Context, request reconcile.Request) (re
...
 
+	// Update the cluster provider type.
+	if r.config.platformType == "" {
+		platformType, err := r.getPlatformType()
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		r.config.platformType = platformType
+	}
+
```
elmiko, line 237: this clause looks fine to me, but could we either move it before the `ensureAutoscalerMonitoring` or before the `UpdateAutoscaler` call, because currently it is breaking up the flow of the `err` return from the `ensureAutoscalerMonitoring`. this isn't a technical problem, but does break up the readability of the error return from the monitoring call.

elmiko (commented):
change looks fine to me, i just have a request about code readability.

maxcao13:
Well... this non-required ovn test is failing because of other cluster operators not being ready (machine-api, authentication) so I don't think these failures are telling us anything, but I could be wrong since I'm new to the clusteroperator space.

PR #348, opened 2025-07-02
NO-JIRA: Remove tombstones yamls

lance5890:
kind of post-cleanup of https://github.com/openshift/cluster-autoscaler-operator/pull/214

address https://github.com/openshift/cluster-version-operator/issues/1207


elmiko:
change lgtm, i wouldn't have expected the failure we are seeing in the test. i don't think it's related to this change though.

/retest

PR #374, opened 2026-05-06
AUTOSCALE-660: add enforce node group min size flag

shyadav10:
Add new field and plumbing for enforce node group min size

## Summary by CodeRabbit

* **New Features**
  * Added an enforceNodeGroupMinSize option to ClusterAutoscaler to enable or disable enforcement of minimum node group sizes (choices: Enabled/Disabled; defaults to Disabled).


pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -98,6 +108,13 @@ type ClusterAutoscalerSpec struct {
 	// +kubebuilder:validation:MaxItems=3
 	// +optional
 	Expanders []ExpanderString `json:"expanders"`
+	// EnforceNodeGroupMinSize enables/disables the `--enforce-node-group-min-size` cluster-autoscaler feature flag.
+	// When enabled, the cluster autoscaler will enforce the minimum size of a node group,
+	// ensuring that the node group never scales below the configured minimum size even if
+	// nodes are deemed unneeded. This is useful for maintaining a baseline capacity in node groups.
+	// Defaults to Disabled.
```
elmiko, line 115: i think we want to add a kubebuilder default validation here, as described in the docs https://book.kubebuilder.io/reference/markers/crd-validation.html


```suggestion
	// Defaults to Disabled.
	// +default:value="Disabled"
```

elmiko (commented):
i think we want a default validation on the crd, also we should update the [example ClusterAutoscaler](https://github.com/openshift/cluster-autoscaler-operator/blob/main/examples/clusterautoscaler.yaml) to include a usage of the new field.

pkg/apis/autoscaling/v1/clusterautoscaler_types.go
shyadav10: done

elmiko (commented):
sorry, forgot to check back here. update looks great, thanks!

/approve
/lgtm

elmiko:
@shyadav10 this just needs to be verified to merge. usually the author will verify based on a test or manually operating the patch. have you tried this out to make sure it works as expected?

shyadav10:
verified

PR #378, opened 2026-06-10
OCPBUGS-87827: only use central TLS profile when the cluster says to

joelsmith:
## Summary by CodeRabbit

* **Chores**
  * Updated OpenShift and Kubernetes client libraries and aggregator to newer versions; removed an unused indirect dependency to streamline the build.
  * Improved TLS profile handling: operator now respects cluster TLS adherence settings and will more promptly reload when cluster TLS policies change, enhancing runtime TLS behavior and security.


joelsmith:
I don't think the failures are related to this PR
/retest

maxcao13:
/lgtm

Are we gonna have e2e tests for this like the other PRs? If so, I think we would have to put that in `cluster-api-actuator-pkg` or `autoscale-tests` if so, since tests don't live in this repo.

joelsmith:
Ah yes, I didn't want to put the tests in the old place since the move is in progress. I will create a card to add a test to the new repo.

elmiko:
i'm not sure what is going on with the failure in regression-clusterinfra-gcp-ipi-autoscaler, but the scale up/down tests in the e2e2-aws-operator appear to be working.

/approve

PR #381, opened 2026-07-30
OCPBUGS-100034: Dump kube RBAC proxy and do metrics endpoint authn & authz in process

joelsmith:
Since kube-rbac-proxy [is being discontinued](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/discontinue_usage_of_kube_rbac_proxy.md) and its use is discouraged, I figured that the easiest way to fix the linked bug is to get rid of kube-rbac-proxy and bring its functionality in process. We have had `FilterProvider` and `SecureServing` for [almost 3 years](https://github.com/kubernetes-sigs/controller-runtime/pull/2407), so I think we're not early to the party. :-D

Note that this change makes it so we use the stock `FilterProvider` `filters.WithAuthenticationAndAuthorization` which implicitly relies upon the existing OpenShift clusterrole `prometheus-k8s` which allows (among other things):
```
- nonResourceURLs:
  - /metrics
  - /metrics/slis
  verbs:
  - get
```


elmiko:
i'm not an expert on the kube-rbac-proxy, but this makes sense to me. 

/lgtm
/approve

joelsmith:
/verified by E2Es
I also verified manually that prometheus is still able to scrape CAO's metrics:
```
$ oc exec -n openshift-monitoring prometheus-k8s-0 -c prometheus -- curl -s 'http://localhost:9090/api/v1/targets' | jq '.data.activeTargets[] | select(.discoveredLabels.__meta_kubernetes_endpointslice_label_k8s_app=="cluster-autoscaler-operator") | .health'
"up"
```
And also saw that there is now only one container in the pod:
```
$ oc get pod -A|grep cluster-autoscaler-operator
openshift-machine-api                              cluster-autoscaler-operator-b5fcd9c4-p4qk8                   1/1     Running     0             42m
```

PR #382, opened 2026-07-31
NO-JIRA: Add /release-chores chore cycle skill

joelsmith:


LucasAndFlores (commented):
Nice Job Joel, I liked a lot the way is configured and looking forward for testing this

PR #383, opened 2026-08-13
OCPBUGS-105277: implement startupTaints

LucasAndFlores:
This PR implements the new field `startupTaints`. The startup-taints should be passed to autoscaler as arguments, and when passed, cluster-autoscaler cluster treats nodes with these taints as unready during scale-up, expecting them to become ready shortly.


pkg/apis/autoscaling/v1/clusterautoscaler_types.go
```diff
@@ -116,6 +116,13 @@ type ClusterAutoscalerSpec struct {
 	// +optional
 	// +kubebuilder:default=Disabled
 	EnforceNodeGroupMinSize *EnforceNodeGroupMinSizeMode `json:"enforceNodeGroupMinSize,omitempty"`
+
+	// Taints that indicate a node is still starting up. The cluster autoscaler treats
+	// nodes with these taints as unready during scale-up, expecting them to become ready shortly.
+	// Each taint must contain only the key.
```
elmiko: we should probably begin this paragraph with the name of the field like the other entries. perhaps something like this:


```suggestion
	// StartupTaints  contains values that indicate the keys of taints that will be on a node that is still starting up. The cluster autoscaler treats
	// nodes with these taints as unready during scale-up, expecting them to become ready shortly.
	// Each taint must contain only the key.
```

pkg/controller/clusterautoscaler/clusterautoscaler.go
```diff
@@ -245,6 +249,15 @@ func AutoscalerArgs(ca *v1.ClusterAutoscaler, cfg *Config) []string {
 		args = append(args, EnableProvisioningRequestsArg.Value(trueFlag))
 	}
 
+	// if feature gate for StartupTaints is enabled, add to the arguments
+	if isFeatureGateEnabled(cfg.FeatureGateAccessor, startupTaintsFGName) {
```
elmiko: since we are bringing this in to GA, i don't think we should make a feature gate for it specifically.

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
```diff
@@ -294,6 +295,214 @@ func TestAutoscalerArgsFromSpec(t *testing.T) {
...
+			expectedMissing: []string{
+				"--scale-down-delay-after-delete",
+				"--scale-down-delay-after-failure",
+				"--max-node-provision-time",
+				"--balance-similar-node-groups",
+				"--ignore-daemonsets-utilization",
+				"--skip-nodes-with-local-storage",
+				"--balancing-ignore-label",
+			},
+		},
```
elmiko: we do a default args test in another test, do we need to repeat it here?

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
```diff
@@ -294,6 +295,214 @@ func TestAutoscalerArgsFromSpec(t *testing.T) {
...
+				[]configv1.FeatureGateName{startupTaintsFGName},
+				[]configv1.FeatureGateName{},
+			)},
+			caFunc: func() *autoscalingv1.ClusterAutoscaler {
+				ca := NewClusterAutoscaler()
+				ca.Spec.StartupTaints = []string{}
+				return ca
+			},
+			expected: []string{},
+		},
```
elmiko: i think we could isolate this test to just these cases.

elmiko (commented):
i think we can remove the featuregate and reduce the testing surface. otherwise looking pretty good.

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
LucasAndFlores: Sorry, my bad here. When I was testing, I copy/paste the above test just to test this specific function. I'll reply you in the other comment

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
LucasAndFlores: Yes, I agree. But since the startupTaints are not behind a feature gate, I think we can include in `TestAutoscalerArgsFromSpec`, because it will come from Spec fields. And I'll isolate this test for feature gate only

pkg/controller/clusterautoscaler/clusterautoscaler_test.go
```diff
@@ -294,6 +315,48 @@ func TestAutoscalerArgsFromSpec(t *testing.T) {
 	}
 }
 
+func TestAutoscalerArgsFeatureGate(t *testing.T) {
```
elmiko, line 318: why add this as a new test? (just curious)

LucasAndFlores: Because this one still in FeatureGate. I commented more [here](https://github.com/openshift/cluster-autoscaler-operator/pull/383#discussion_r3782763130)

LucasAndFlores: But I don't get it, do you want me to remove it or keep it?

elmiko (commented):
/approve
/lgtm

PR #384, opened 2026-08-17
OCPBUGS-103824: Bump x/net package in release-4.20

jkaurredhat:
govulncheck confirms Symbol Results. Operator calls vulnerable HTTP/2 transport functions through Kubernetes API client chain. Vulnerable transport.go present in vendor. Attack scenario: malicious HTTP/2 server sending SETTINGS_MAX_FRAME_SIZE=0 causes infinite loop. Mitigating: operator connects to trusted API server.


go.mod
```diff
@@ -86,3 +87,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.6.0 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace golang.org/x/net => github.com/openshift-sustaining/net v0.50.0-sec.3
```
LucasAndFlores, line 91: @jkaurredhat Out of curiosity, why do we need this? I can see that we bumped the version for this library

jkaurredhat: @LucasAndFlores So the version which is there is upstream for this fix requires a golang bump and do not work with go1.24 hence needed the fork with the patch.

PR #386, opened 2026-08-17
OCPBUGS-104038:Bump x/net package in release-4.21

jkaurredhat:
govulncheck confirms Symbol Results. Operator calls vulnerable HTTP/2 transport functions through Kubernetes API client chain. Vulnerable transport.go present in vendor. Attack scenario: malicious HTTP/2 server sending SETTINGS_MAX_FRAME_SIZE=0 causes infinite loop. Mitigating: operator connects to trusted API server.

Using fix from here : https://github.com/openshift-sustaining/net/releases


jkaurredhat:
/verified by regression-clusterinfra-gcp-ipi-autoscaler can be ignored

PR #395, opened 2026-08-18
OCPBUGS-111092: update version to properly detect errors

elmiko:
This change is being proposed to fix a condition where a malformed version string during build time can result in a panic during execution time. The CAO will now interrogate the compiled version to ensure that it won't panic during execution, and update the displayed version accordingly.

In cases where the version follows semver, it will use that version. If the version string does not adhere to semver, then the version will be `0.0.0-<version string>`. If the version string produces errors then the version will be reported as `0.0.0-unable-to-set-version`.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -1,3 +1,19 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2026 Red Hat, Inc.
+ *
+ */
 package version
 
 import (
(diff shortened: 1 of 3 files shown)


pkg/version/version.go
```diff
@@ -7,14 +23,34 @@ import (
...
 	// Version is semver representation of the version.
-	Version = semver.MustParse(strings.TrimLeft(Raw, "v"))
+	Version = mustParse(Raw)
 
 	// String is the human-friendly representation of the version.
-	String = fmt.Sprintf("cluster-autoscaler-operator %s", Raw)
+	String = fmt.Sprintf("cluster-autoscaler-operator %s", Version)
 )
+
+func mustParse(v string) semver.Version {
```
LucasAndFlores, line 42: I suggest improve how the function is written. I'd write like this: 
```go
	version, err := semver.Parse(strings.TrimLeft(v, "v"))
	if err != nil {
		// version did not parse cleanly, append it to the zero value
		appendedVersion, err := semver.Parse(fmt.Sprintf("0.0.0-%s", v))
		if err != nil {
			// still having an issue making a clean version, use a known value
			// this shouldn't happen in practice
			return semver.MustParse(errorVersion)
		}

		return appendedVersion
	}

	return version
```
IMHO, it makes easier to read and eliminates the `else` expressions, which can turn the readability a little bit harder. I tried this rewrite locally with the tests that you wrote and it is passing.

elmiko: yeah, functionally the same. i think i left the `else` there because i had just done it on the other clause.

updating

elmiko:
updated to:
* clean up mustParse function
* fix syntax error in Makefile

pkg/version/version.go
LucasAndFlores: But do you want to keep the first clause? I'm good if you want

LucasAndFlores:
/approve
/lgtm 
Thanks for the changes :slightly_smiling_face:

PR #396, opened 2026-08-19
OCPBUGS-111996:[release-4.20]: implement startupTaints

LucasAndFlores:
This PR implements the new field startupTaints. The startup-taints should be passed to autoscaler as arguments, and when passed, cluster-autoscaler cluster treats nodes with these taints as unready during scale-up, expecting them to become ready shortly.


elmiko:
we are not planning a backport for now

/close

PR #399, opened 2026-08-20
AUTOSCALE-899: add claude and contributing files

LucasAndFlores:
Since we already have it ` README.md` and ` .coderabbit.yaml`, I only added `AGENTS.md` and `CONTRIBUTING.md`. I also added two new points that I think is relevant for AI workflow, which is verification and code conventions. These new points are relevant because we can ""force"" AI to follow the way that we write the code in this repo.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,298 @@
+# AGENTS.md — openshift/cluster-autoscaler-operator
+
+This file provides AI-specific guidance for working in the Cluster Autoscaler Operator repository. For contribution guidelines, see [CONTRIBUTING.md](CONTRIBUTING.md).
+
+## Project Overview
+
+This repo is the **Cluster Autoscaler Operator** for OpenShift — a standalone operator (`github.com/openshift/cluster-autoscaler-operator`) that manages the lifecycle of the Kubernetes Cluster Autoscaler on OpenShift clusters.
+
+The operator is deployed by the Cluster Version Operator (CVO) into the `openshift-machine-api` namespace. It can also run on vanilla Kubernetes clusters with machine-api components available.
+
+### Single Binary
+
+This repo produces one binary:
+
+| Binary | Source | Purpose |
+|--------|--------|---------|
+| `cluster-autoscaler-operator` | `cmd/manager/main.go` | The operator. Watches ClusterAutoscaler/MachineAutoscaler CRs and manages the cluster-autoscaler Deployment, monitoring, network policies, webhooks, and CVO status. |
+
+### CRDs
+
+Both CRDs belong to API group `autoscaling.openshift.io`.
+
+| CRD | API Version | Scope | Purpose |
+|-----|-------------|-------|---------|
+| ClusterAutoscaler | `autoscaling.openshift.io/v1` | Cluster | Singleton (`"default"`) configuring the cluster-wide autoscaler: resource limits, scale-down/up settings, expanders, startup taints, GPU limits |
+| MachineAutoscaler | `autoscaling.openshift.io/v1beta1` | Namespaced | Defines min/max replica bounds for a specific MachineSet target |
+
+## Repository Structure
+
+```
+cmd/manager/                    # Single binary entry point
+pkg/
+  apis/autoscaling/v1/          # ClusterAutoscaler CRD types + deepcopy
+  apis/autoscaling/v1beta1/     # MachineAutoscaler CRD types + deepcopy
+  controller/clusterautoscaler/ # ClusterAutoscaler reconciler, monitoring, network policies, validation
+  controller/machineautoscaler/ # MachineAutoscaler reconciler, MachineTarget management, validation
+  operator/                     # Operator bootstrap, env-var config, CVO status reporter, webhook config
+  util/                         # Shared utilities (validation, TLS profile, leader election, units)
+  version/                      # Build-time version string
+install/                        # CVO install manifests (CRDs, RBAC, Deployment, etc.), numbered for ordering
+hack/                           # Build and CI scripts (test runner, CRD generation, formatting, vendoring)
+test/helpers/                   # Shared test helper utilities (TestDeployment, TestClusterOperator builders)
+examples/                       # Sample ClusterAutoscaler and MachineAutoscaler CR YAMLs
+docs/dev/                       # Developer docs (metrics scraping)
+docs/user/                      # User docs (Prometheus alerts)
+.claude/commands/               # Claude Code slash commands (release-chores)
+```
+
+## Relationship to Upstream kubernetes-autoscaler
+
+This operator does **not** contain any upstream autoscaler code. It is a standalone operator that manages a Deployment of the upstream cluster-autoscaler.
+
+- The autoscaler operand lives in the [kubernetes-sigs/cluster-autoscaler](https://github.com/kubernetes-sigs/cluster-autoscaler).
+- The operator's job: Runs the upstream autoscaler image.
+- The autoscaler image is injected via the `CLUSTER_AUTOSCALER_IMAGE` env var (set by CVO from `install/image-references`).
+- When working on this repo, you rarely need to touch upstream autoscaler code. Changes to autoscaler behavior are done via the CR spec fields here.
+
+## Architecture: What Is Not Obvious
+
+### Operator-Manages-Deployment Pattern
+
+The operator creates and manages a separate Deployment (`cluster-autoscaler-<name>`) in the same namespace. It does not run the autoscaler in-process. The reconcile loop in `pkg/controller/clusterautoscaler/clusterautoscaler_controller.go` handles the full lifecycle: create, update, and delete the Deployment when the CR changes.
+
+The operator also creates owned secondary resources:
+- **Monitoring**: Service (port 8085), ServiceMonitor, PrometheusRule (4 alerts: UnschedulablePods, NotSafeToScale, CPULimitReached, MemoryLimitReached) — see `monitoring.go`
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,209 @@
...
+Every change should be closely scrutinized for bugs. Review changes from multiple angles:
+
+- **API correctness**: Do changes to CRD types in `pkg/apis/` follow naming conventions, include proper validation markers, and maintain backward compatibility?
+- **Reconciliation logic**: Is error handling correct? Are status updates and event recording consistent?
+- **ClusterOperator status**: Could this change affect the operator's Available, Progressing, or Degraded conditions?
+- **RBAC and security**: Are there new permissions required? Are changes to `install/03_rbac.yaml` or RBAC-related code properly scoped?
+- **Webhook validation**: Are validating webhook changes correct and complete?
+- **Generated artifacts**: Are CRD YAMLs and deepcopy files regenerated and up to date?
+
+## Commit Convention
```
linkvt: Im not sure if we follow conventional commits in our repos, I didn't have the feeling that this is the case? I'm fine with adding it but not sure if we want to enforce it.

AGENTS.md
```diff
@@ -0,0 +1,303 @@
...
+
+**Always write tests for new code.** Every code change must have unit test coverage.
+
+**Prefer adding cases to existing test files** over creating new test files. Each source file has a corresponding `_test.go` file (e.g., `clusterautoscaler.go` / `clusterautoscaler_test.go`). Add new test cases to the existing table-driven tests when the new behavior fits an existing test function. Only create a new test function or file when the behavior being tested is genuinely different from existing tests.
+
+**Table-driven tests with subtests** are the dominant pattern:
+
+```go
+testCases := []struct {
+    label    string
```
linkvt, line 182: Probably very subjective but I prefer table tests using a map, e.g. as documented here: https://go.dev/wiki/TableDrivenTests#using-a-map-to-store-test-cases

Mostly because of the mentioned benefits:
> One advantage of using maps is that the “name” of each test can simply be the map index.
> 
> More importantly, map iteration order isn’t specified nor is it even guaranteed to be the same from one iteration to the next. This ensures that each test is independent of the others and that testing order doesn’t impact results.


Also not sure about the need to specify the test conventions as detailed as below considering my other comment about LLM opinions/redundant infos but very subjective so feel free to ignore :)

AGENTS.md
```diff
@@ -0,0 +1,303 @@
...
+
+### Constants Over Inline Strings
+
+Always define named constants for annotation keys, label names, resource names, flag names, default config values, and reason strings. Never use inline string literals for values that are referenced in more than one place or that represent a contract (annotations, labels, API field names).
+
+Constants are grouped into `const` blocks organized by topic. Exported constants get godoc comments. Private constants (lowercase) are used for values that don't leave the package.
+
+This codebase also uses typed string constants when a group of related values needs formatting methods. For example, `AutoscalerArg` in `pkg/controller/clusterautoscaler/clusterautoscaler.go` is a `type AutoscalerArg string` with methods like `String()`, `Value()`, `Range()` that ensure consistent CLI flag formatting.
+
+### Methods vs Functions
```
linkvt, line 154: Not sure if we want to include advice like this as it is IMO the default/best practice?

When I was creating the files for the [HTTP Addon](https://github.com/openshift/kedacore-http-add-on/pull/9) the LLM I used also included lots of infos that were common sense or best practice bloating the files quite a lot.
I think I spent most of my time removing things that were redundant or based on LLM opinions.

But as I said just my opinion and I'm not contributing to this repo so I don't know what is important to the people maintaining it :)

CONTRIBUTING.md
LucasAndFlores: I agree with you, I'll remove it. I'll keep using it (and I like a lot), but I also don't see so much in other repos, so it doesn't make sense to keep it

AGENTS.md
LucasAndFlores: I liked a lot this map option, thanks for sharing and I'll try it in a future. But I think it is good to keep this, since I already have some experiences where the LLM didn't follow the convention, that's why I added.  It is more about a "safeguard" the development with the LLM

linkvt: Sounds good, thanks for explaining :)

AGENTS.md
LucasAndFlores: Yeah, about the typed string constant I agree with you. But about how to use consts or strings, I think it is interesting.

CONTRIBUTING.md
```diff
@@ -0,0 +1,187 @@
+# Contributing to Cluster Autoscaler Operator
+
+This document covers contribution guidelines for the [Cluster Autoscaler Operator](https://github.com/openshift/cluster-autoscaler-operator), which manages deployments of the OpenShift Cluster Autoscaler. For architecture and usage details, see the [README](README.md).
+
+## Related Resources
+
+| Resource | Link |
+|----------|------|
+| Upstream Cluster Autoscaler | [openshift/kubernetes-autoscaler](https://github.com/openshift/kubernetes-autoscaler) |
```
joelsmith: This is the downstream CA link (which is good to have)
Upstream CA is https://github.com/kubernetes/autoscaler

AGENTS.md
```diff
@@ -0,0 +1,299 @@
+# AGENTS.md — openshift/cluster-autoscaler-operator
+
+This file provides AI-specific guidance for working in the Cluster Autoscaler Operator repository. For contribution guidelines, see [CONTRIBUTING.md](CONTRIBUTING.md).
+
+## Project Overview
+
+This repo is the **Cluster Autoscaler Operator** for OpenShift — a standalone operator (`github.com/openshift/cluster-autoscaler-operator`) that manages the lifecycle of the Kubernetes Cluster Autoscaler on OpenShift clusters. It does not contain the autoscaler itself; instead, it creates and manages a Deployment that runs the upstream autoscaler image.
```
joelsmith: I think it's confusing to talk about upstream in this context. I'd say it would be better to say either "operand" (i.e. the thing the operator operates) or "OpenShift".

AGENTS.md
```diff
@@ -0,0 +1,299 @@
...
+docs/dev/                       # Developer docs (metrics scraping)
+docs/user/                      # User docs (Prometheus alerts)
+.claude/commands/               # Claude Code slash commands (release-chores)
+```
+
+## Relationship to Upstream kubernetes-autoscaler
+
+This operator does **not** contain any upstream autoscaler code. It is a standalone operator that manages a Deployment of the upstream cluster-autoscaler.
+
+- The upstream autoscaler lives in the [openshift/kubernetes-autoscaler](https://github.com/openshift/kubernetes-autoscaler) fork (a downstream fork of `kubernetes/autoscaler`).
```
joelsmith: Maybe "The autoscaler operand lives..."

AGENTS.md
```diff
@@ -0,0 +1,299 @@
...
+- The operator's job: translate the `ClusterAutoscaler` CR spec into command-line arguments and create/update a Deployment that runs the upstream autoscaler image.
```
joelsmith: another upstream. This one I think you could just remove, i.e. "runs the autoscaler"

AGENTS.md
```diff
@@ -0,0 +1,299 @@
...
+- The autoscaler image is injected via the `CLUSTER_AUTOSCALER_IMAGE` env var (set by CVO from `install/image-references`).
+- CR spec fields map to `--flag=value` arguments in `pkg/controller/clusterautoscaler/clusterautoscaler.go` (see the `AutoscalerArgs()` function).
+- When working on this repo, you rarely need to touch upstream autoscaler code. Changes to autoscaler behavior are done via the CR spec fields here.
```
joelsmith, line 56: another `upstream` to remove.

AGENTS.md
```diff
@@ -0,0 +1,299 @@
...
+
+Ownership is tracked via the `autoscaling.openshift.io/machineautoscaler` annotation and a finalizer (`machinetarget.autoscaling.openshift.io`). When a MachineAutoscaler's target changes, the old target's annotations are cleaned up before the new target is configured.
+
+### Environment Variable Configuration
+
+All operator configuration comes from environment variables, parsed in `pkg/operator/config.go`. There are no config files or ConfigMaps.
+
+| Variable | Default | Purpose |
+|----------|---------|---------|
+| `WATCH_NAMESPACE` | `openshift-machine-api` | Namespace to watch for CRs |
```
joelsmith, line 92: Do you think we could do without this table? I think the above reference is probably enough since an agent or human reading this can read the comments in the file which in the variable declaration section are quite extensive. My concern is that it would be a pain to keep this table in sync as options are added, removed or changed.

joelsmith (commented):
I have a few small suggestions. Otherwise, looks good to me. I'll defer to @linkvt to give the final `/lgtm` once he is happy with it since he did the initial review. Thanks!

/approve

CONTRIBUTING.md
LucasAndFlores: I think I'll do even better, maybe use https://github.com/kubernetes-sigs/cluster-autoscaler instead of this one. Since it will be migrated in the future, I think it is worth it to reference the new one

AGENTS.md
LucasAndFlores: I agree, I'll keep the reference to the file only. Way easier to keep updated, thanks for the comment

linkvt:
/lgtm

Thanks, looks good to me and also thanks Joel for the helpful review 🙏

PR #402, opened 2026-09-01
AUTOSCALE-709: implement conditions and authoritative API

LucasAndFlores:


pkg/controller/machineautoscaler/condition.go
```diff
@@ -0,0 +1,31 @@
+package machineautoscaler
+
+import (
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func createOrUpdateMaCondition(conditions []metav1.Condition, newCondition metav1.Condition) []metav1.Condition {
```
elmiko, line 9: i wonder if we should use the `SetCondition` function from our library-go package?

https://github.com/openshift/library-go/blob/master/pkg/operator/v1helpers/helpers.go#L512-L530

pkg/controller/machineautoscaler/machineautoscaler_controller.go
```diff
@@ -585,3 +685,53 @@ func objectReference(ref v1beta1.CrossVersionObjectReference) *corev1.ObjectRefe
 
 	return obj
 }
+
+func validateAuthoritativeAPI(authoritativeAPIToGVK map[machinev1.MachineAuthority]schema.GroupVersionKind, ref *corev1.ObjectReference, target *MachineTarget) error {
+	val, exists, err := unstructured.NestedString(target.Object, "status", "authoritativeAPI")
```
elmiko: i wonder if we would want to check the `spec.authoritativeAPI` for any reason?

pkg/controller/machineautoscaler/machineautoscaler_controller.go
```diff
@@ -585,3 +685,53 @@ func objectReference(ref v1beta1.CrossVersionObjectReference) *corev1.ObjectRefe
...
+
+	if err != nil {
+		klog.Errorf("Error getting authoritativeAPI from target.Object: %v", err)
+		return err
+	}
+
+	if !exists {
+		return ErrAuthoritativeTypeNotExist
```
elmiko, line 698: we might need to accept some form of backward compatibility, i'm guessing that the authoritative field would always be there, but do we have any scenarios where it would legitimately not be there?

pkg/controller/machineautoscaler/machineautoscaler_controller.go
```diff
@@ -585,3 +685,53 @@ func objectReference(ref v1beta1.CrossVersionObjectReference) *corev1.ObjectRefe
...
+
+	if authoritativeAPI == "" {
+		return ErrAuthoritativeTypeInvalid
+	}
+
+	if authoritativeAPI == machinev1.MachineAuthorityMigrating {
+		return ErrMigratingAuthoritativeType
+	}
+
+	gvk, exists := authoritativeAPIToGVK[authoritativeAPI]
```
elmiko, line 711: is `authoritativeAPIToGVK` a cache?

elmiko (commented):
i gave a quick review, have some questions.

pkg/controller/machineautoscaler/condition.go
LucasAndFlores: We could, but I think if we use this function we will keep only one condition, since it uses `Type` field to find and update. [In the rfe](https://github.com/openshift/enhancements/blob/master/enhancements/cluster-api/cluster-autoscaler-integration-with-openshift-cluster-api.md#machineautoscaler-status-changes), the `Type` field is the same for failure or success. If this is the expected behaviour, I can change it. But I understood that we should keep organized using `Reason`.

pkg/controller/machineautoscaler/machineautoscaler_controller.go
LucasAndFlores: Maybe I misunderstood this one. Because in the example [on the RFE](https://github.com/openshift/enhancements/blob/master/enhancements/cluster-api/cluster-autoscaler-integration-with-openshift-cluster-api.md#machineautoscaler-spec-changes), I understood we would use `status.authoritativeAPI`, but it states this at the end: 
> To further prevent users from misconfiguring MachineAutoscalers or MachineSets, a validating admission policy will be installed to prevent a MachineSet from changing its authoritative ownership (.spec.authoritativeAPI) while a MachineAutoscaler references that MachineSet. This will help ensure that no MachineAutoscaler is referencing the wrong scalable resource.

I'll change it

pkg/controller/machineautoscaler/machineautoscaler_controller.go
LucasAndFlores: We are checking only if the feature gate is enabled, so it "should" be there. But we are checking if the field is not there or empty in the line that you pointed out (697) or if the field `spec.authoritativeAPI` is empty or misconfigured (between the lines 701-705). I don't know if we have any other possible way to check this.

pkg/controller/machineautoscaler/machineautoscaler_controller.go
LucasAndFlores: Not really, authoritativeAPIToGVK is a mapper, that maps `MachineAuthority` as key to groupVersionKind. If we have `spec.authoritativeAPI` value as `MachineAPI` the groupVersionKind should be`{Group: "machine.openshift.io", Version: "v1beta1", Kind: "MachineSet"}`. This is a way that I found it should make the map easier and explicit in the code.

pkg/controller/machineautoscaler/machineautoscaler_controller.go
LucasAndFlores: Or should we check both? Now I'm confused :smile:

pkg/controller/machineautoscaler/condition.go
elmiko: `SetCondition` does a number of things:
* if `conditions` is empty is creates a new slice
* if `conditions` contains something with the same `.Type` as `newCondition` it updates
* if `conditions` does not contain `newCondition` it is added
* it preserves other entries in the `conditions` which do not match

we shouldn't have 2 conditions with the same `Type`, we want to be updating the `Ready` type condition when there is something that prevents operation.

pkg/controller/machineautoscaler/machineautoscaler_controller.go
elmiko: i think you did it correctly, we should probably only be checking the `.status` to prevent the possibility of a race condition with checking the `.spec` field.

pkg/controller/machineautoscaler/machineautoscaler_controller.go
```diff
@@ -585,3 +685,53 @@ func objectReference(ref v1beta1.CrossVersionObjectReference) *corev1.ObjectRefe
 
 	return obj
 }
+
+func validateAuthoritativeAPI(authoritativeAPIToGVK map[machinev1.MachineAuthority]schema.GroupVersionKind, ref *corev1.ObjectReference, target *MachineTarget) error {
+	val, exists, err := unstructured.NestedString(target.Object, "spec", "authoritativeAPI")
```
elmiko: given our discusion, i think this should be `status` instead of `spec`.

pkg/controller/machineautoscaler/machineautoscaler_controller.go
```diff
@@ -355,6 +420,34 @@ func (r *Reconciler) HandleTargetChange(ma *v1beta1.MachineAutoscaler) error {
...
+				return err
+			}
+		}
+
+		condition := createCondition(metav1.ConditionTrue, machineAutoscalerReadyType, machineAutoscalerReadyReason, machineAutoscalerReadyMessage)
+		err = r.updateMachineAutoscalerConditions(ma, condition)
+		if err != nil {
+			klog.Errorf("Error updating machine autoscaler conditions: %v", err)
+			return err
+		}
```
elmiko, line 470: it seems like we might want to put this in an `else` block, otherwise we could update the conditions twice.

pkg/controller/machineautoscaler/machineautoscaler_controller.go
```diff
@@ -234,6 +279,26 @@ func (r *Reconciler) Reconcile(_ context.Context, request reconcile.Request) (re
...
+				return reconcile.Result{}, err
+			}
+		}
+
+		condition := createCondition(metav1.ConditionTrue, machineAutoscalerReadyType, machineAutoscalerReadyReason, machineAutoscalerReadyMessage)
+		err = r.updateMachineAutoscalerConditions(ma, condition)
+		if err != nil {
+			klog.Errorf("Error updating machine autoscaler conditions: %v", err)
+			return reconcile.Result{}, err
+		}
```
elmiko, line 319: we might also want this in an `else` block to prevent double writing the condition.

install/03_rbac.yaml
```diff
@@ -522,7 +522,7 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: cluster-autoscaler-operator
-    namespace: openshift-machine-api
+    namespace: openshift-cluster-api
```
elmiko: good catch, our service account is scoped to the openshift-machine-api namespace, we need to either update that to cover both namespaces or add another service account.

pkg/controller/machineautoscaler/condition.go
LucasAndFlores: Ok, I updated it this one. Thanks for catching

pkg/controller/machineautoscaler/machineautoscaler_controller.go
LucasAndFlores: I also revert back to `.status` instead of `.spec`

pkg/controller/machineautoscaler/machineautoscaler_controller.go
LucasAndFlores: Updated it

pkg/controller/machineautoscaler/machineautoscaler_controller.go
LucasAndFlores: Actually, I think I can add only a `return nil` right after the `err != nil`. I'd prefer like this, but let me know WDYT

pkg/controller/machineautoscaler/machineautoscaler_controller.go
LucasAndFlores: Same here, I think I'd prefer to use `return reconcile.Result{}, nil` right after `err != nil`

pkg/controller/machineautoscaler/machineautoscaler_controller.go
```diff
@@ -101,6 +136,16 @@ func (r *Reconciler) AddToManager(mgr manager.Manager) error {
 		return err
 	}
 
+	if r.isClusterAPIIntegrationEnabled {
+		r.authoritativeAPIToGVK = map[machinev1.MachineAuthority]schema.GroupVersionKind{
+			machinev1.MachineAuthorityMachineAPI: r.config.SupportedTargetGVKs[0],
+			machinev1.MachineAuthorityClusterAPI: clusterAPIGVK,
+		}
```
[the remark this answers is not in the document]
LucasAndFlores: This was a good one, I'll implement it

install/03_rbac.yaml
LucasAndFlores: Ok, I tried another thing in this file. Instead of using `ClusterRoleBinding` without a namespace, I tried to use a `RoleBinding` scoped in `openshift-cluster-api` namespace. This will help us to reuse mostly everything and keep these role only for `openshift-cluster-api` namespace. Let me know WDYT

pkg/controller/machineautoscaler/machineautoscaler_controller.go
```diff
@@ -101,6 +139,16 @@ func (r *Reconciler) AddToManager(mgr manager.Manager) error {
...
 	}
 
+	if r.isClusterAPIIntegrationEnabled {
+		r.authoritativeAPIToGVK = map[machinev1.MachineAuthority]schema.GroupVersionKind{
+			machinev1.MachineAuthorityMachineAPI: machineAPIGVK,
+			machinev1.MachineAuthorityClusterAPI: clusterAPIGVK,
+		}
+
+		// if the feature gate is enabled, we should append the clusterAPIGVK to be watched and checked
+		r.config.SupportedTargetGVKs = append(r.config.SupportedTargetGVKs, clusterAPIGVK)
```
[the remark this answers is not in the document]
LucasAndFlores: Point for me where the cache and watches are using the namespace

install/03_rbac.yaml
LucasAndFlores: I'm starting to think that this change from `ClusterRoleBinding` to `RoleBinding` in the namespace `openshift-cluster-api` could be the cause of failure in the e2e tests. The roleBinding was in the namespace `openshift-cluster-api`, which maybe was not created and resulting the error like this: 
```
level=error msg=failed to initialize the cluster: Could not update rolebinding "openshift-cluster-api/cluster-autoscaler-operator-clusterapi" (355 of 1052): resource may have been deleted: timed out waiting for the condition
```
I tried to convert to `ClusterRoleBinding` and let's see how the tests behaves. It is not a 100% wrong approach, but will give a broader permission (cluster level) instead of namespace scope. 
WDYT? I'm not a 100% sure of this, but it looks like is the fix

pkg/controller/machineautoscaler/machineautoscaler_controller.go
elmiko, line 292: if we error when getting the target scale ref, let's add a condition here as well. with a message like "cannot find targetScaleRef"

pkg/controller/machineautoscaler/machineautoscaler_controller.go
```diff
@@ -355,6 +440,36 @@ func (r *Reconciler) HandleTargetChange(ma *v1beta1.MachineAutoscaler) error {
 		}
 	}
 
+	if r.isClusterAPIIntegrationEnabled {
+		newTarget, err := r.GetTarget(newTargetRef)
+		if err != nil {
```
elmiko, line 445: let's also add a condition here, same message as the other "cannot find targetScaleRef"

pkg/controller/machineautoscaler/machineautoscaler_controller.go
```diff
@@ -355,6 +440,36 @@ func (r *Reconciler) HandleTargetChange(ma *v1beta1.MachineAutoscaler) error {
...
+			errMsg := fmt.Sprintf("Error getting new target: %v", err)
+			r.recorder.Eventf(ma, newTargetRef, corev1.EventTypeWarning, "FailedGetNewTarget", "GetNewTarget", "Error fetching new target: %v", err)
+			klog.Errorf("%s: %s", maName, errMsg)
+			return err
+		}
+
+		err = validateAuthoritativeAPI(r.authoritativeAPIToGVK, newTargetRef, newTarget)
```
elmiko, line 452: something i'm concerned about here is that we could potentially have a Cluster API MachineSet as the target ref, and in that case it won't have a `.status.authoritativeAPI`. we will need to check and see if there is a Machine API MachineSet with the same name, and then inspect that MachineSet to see if it has the `authoritativeAPI` field set.

if no corresponding Machine API MachineSet exists, then we should allow the target ref if it exists. in this manner we can support Cluster API resources that were created without a corresponding Machine API resource.

i see a test for this state, but i'm curious if we are supporting it properly?

pkg/controller/machineautoscaler/validator_test.go
```diff
@@ -14,15 +14,15 @@ const (
 	TestMaxReplicas = 8
 )
 
-func NewMachineAutoscaler() *autoscalingv1beta1.MachineAutoscaler {
+func NewMachineAutoscaler(namespace string) *autoscalingv1beta1.MachineAutoscaler {
```
elmiko, line 17: i don't think we need to update the namespace on the MachineAutoscaler, they should stay in the `openshift-machine-api` namespace.

elmiko (commented):
this is looking good, i have a question about the state where we only have a Cluster API MachineSet without a corresponding Machine API MachineSet, and i think we should add another error condition for when the `getTarget` function returns an error.

pkg/controller/machineautoscaler/machineautoscaler_controller.go
LucasAndFlores: Could you please provide the example of condition? It can be something like the yaml, not the code itself. Just to speed up the review

pkg/controller/machineautoscaler/validator_test.go
LucasAndFlores: But why? Can't we have machineautoscaler only in this namespace? Because this function is only used for testing.

pkg/controller/machineautoscaler/machineautoscaler_controller.go
elmiko: i think something like this would work:
```
type: Ready
status: False
reason: ScaleTargetRefNotFound
message: "cannot find scaleTargetRef, check logs for error messages"
```

elmiko (commented):
@LucasAndFlores did you end up removing the rbac changes?

pkg/controller/machineautoscaler/validator_test.go
elmiko: oh gotcha, if it's only for testing that's fine. i missed that detail XD

